PR #28, opened 2015-02-08
Fix bower.json Format

felixbuenemann:
This fixes some issues with the bower.json:
* authors has wrong format
* homepage has wrong format (missing scheme)
* added repository info

More info: https://github.com/bower/bower.json-spec

The malformed homepage url currently breaks packaging with https://rails-assets.org.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -18,10 +18,14 @@
     "console"
   ],
   "authors": [
-    "zloirock.ru"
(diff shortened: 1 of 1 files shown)


felixbuenemann:
Thanks @zloirock, can you push a 0.5.1 release for this?

PR #42, opened 2015-03-13
MIT license?

loganfsmyth:
Your `package.json` says the license is MIT, but I generally find it's much easier quickly glance at a file in the root.


zloirock:
Thanks :)

PR #44, opened 2015-03-22
proposal: Reflect.getOwnPropertyDescriptors

Jabher:
It is not in specification, however, I think it will be there soon.
Reflect represents all reflection methods that were stored in Object (.defineProperty, .hasOwnProperty etc), but there is one method in ES7 proposal that is stored only in Object: getOwnPropertyDescriptors.
Thinking logically, I would predict Reflect.getOwnPropertyDescriptors method that will be 100% similiar to Object. one.

I've just copied the method from Object & tests.


zloirock:
Thanks for the suggestion, but I'm not sure about it. Not all `Object` methods has  `Reflect` equal, only [internal methods](http://people.mozilla.org/~jorendorff/es6-draft.html#sec-ordinary-and-exotic-objects-behaviours). For examle, `Object.keys`, `Object.getOwnPropertyNames` and `Object.getOwnPropertySymbols` covered by `[[OwnPropertyKeys]]` -> `Reflect.ownKeys`. [In proposal](https://gist.github.com/WebReflection/9353781) `Object.getOwnPropertyDescriptors` uses `[[OwnPropertyKeys]]` and `[[GetOwnProperty]]` (`Reflect.getOwnPropertyDescriptor`). Also, all `Reflect` methods should throw error on primitive.

Jabher:
That sounds reasonable.
However, I think that having `Reflect.getOwnPropertyDescriptor` and not having `Reflect.getOwnPropertyDescriptors` in ES spec (not core-js, but spec itself) is some kind of mistake.

zloirock:
I think biggest mistake was in removing `Reflect.hasOwn` from spec :) In ESDiscuss recently opened a [related discussion](https://esdiscuss.org/topic/reflect-getownpropertysymbols).

PR #49, opened 2015-04-08
fixed $.task for IE11 (had a missing context)

bloodyowl:
IE11 threw a `Invalid Calling Object` error because `setImmediate` was called without `window` as context.


zloirock:
Please, show me broken example, I don't understand the problem. This line should not be called in IE11.

bloodyowl:
sorry, my bad, didn't read the code correctly. I'll try to reproduce the case with the least amount of code possible.

closing this in the meantime.

PR #52, opened 2015-04-23
Hook up the standard `npm test` command to run the core-js test suite.

cscott:


zloirock:
Ok, thanks :)

zloirock:
Hmm, interesting, not thought will be so many errors in PhantomJS. Most - `this` in strict mode case.

cscott:
There are pull requests waiting for you that fix all but two.​

PR #53, opened 2015-04-23
Fix strict mode when testing with PhantomJS 1.x.

cscott:
PhantonJS 1.x (and old versions of webkit, Safari, etc) don't properly
handle "use strict" directives if they are nested in functions more than
two deep.  Add a top-level "use strict" inside the first nested function
scope in order to prevent this bug from affecting core-js.

### DIFF
--- a/build/build.ls
+++ b/build/build.ls
@@ -97,6 +97,7 @@ module.exports = ({modules, blacklist, library}, next)-> let @ = modules.turn ((
   next """
     #banner
     !function(undefined){
+    'use strict';
     var __e = null, __g = null;
     
     #script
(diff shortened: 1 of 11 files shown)


zloirock:
Thanks! The problem is that this will only work with UMD wrapper, not with CommonJS entry points. Before `0.7` `"use strict"` was only in this place. Although the correct behavior in strict mode for older browsers is not critical.

cscott:
Yeah, it's not ideal, but it does a little bit better.  I think the
tests-library.js will test the CommonJS entry points, and thus catch if we
miss a function-level "use strict"?  Older browsers should use the UMD
wrapper, I guess, if they care about strict mode behavior.​

zloirock:
>  I think the tests-library.js will test the CommonJS entry points, and thus catch if we
miss a function-level "use strict"?

No. It's tests for version w/o global namespace pollution. CommonJS entry points basely tested in `commonjs.ls` and in some tests not added to repo.

PR #54, opened 2015-04-23
Ensure that we replace Date#toISOString if native implementation is broken

cscott:
The Date#toISOString function in PhantonJS 1.x (and old versions of
webkit, Safari, etc) returns "-01" in the millisecond field in some
cases.  Test for this and ensure that we override the native
implementation in this case.

### DIFF
--- a/modules/es5.js
+++ b/modules/es5.js
@@ -240,7 +240,11 @@ function lz(num){
   return num > 9 ? num : '0' + num;
 }
 // 20.3.4.36 / 15.9.5.43 Date.prototype.toISOString()
-$def($def.P, 'Date', {toISOString: function(){
+function brokenDate() {
+  // PhantomJS and old webkit had a broken Date implementation.
+  return new Date(-5e13 - 1).toISOString() !== '0385-07-25T07:06:39.999Z';
+}
+$def($def.P + $def.F * brokenDate(), 'Date', {toISOString: function(){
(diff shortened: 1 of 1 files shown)


zloirock:
Merge it, but not sure. See, for example, 2 last `Date#toISOString` tests - until recently, behavior this method differed in actual browsers.

cscott:
It's not the leading - which is the issue: phantomJS is generating
"0385-07-25T07:06:39.-01Z"  It's the last four characters which are crazy.
We should patch that away.

We could come up with a more precise test for the "-01" part if you
discover that this is inadvertently patching correct implementations.
​

modules/es5.js
```diff
@@ -240,7 +240,11 @@ function lz(num){
   return num > 9 ? num : '0' + num;
 }
 // 20.3.4.36 / 15.9.5.43 Date.prototype.toISOString()
-$def($def.P, 'Date', {toISOString: function(){
+function brokenDate() {
+  // PhantomJS and old webkit had a broken Date implementation.
+  return new Date(-5e13 - 1).toISOString() !== '0385-07-25T07:06:39.999Z';
```
zloirock: It throws in ES3 browsers - `toISOString` is absent =\

cscott: Ah, yes, brokenDate should obviously return true in ES3.

PR #55, opened 2015-04-23
Ensure that "new RegExp" always creates a new object.

cscott:
This is a fix for PhantomJS (and older versions of webkit/Safari).


zloirock:
Thanks :)

PR #56, opened 2015-04-23
Check whether Object.freeze is stubbed out before using it in tests.

cscott:
In PhantomJS (and older webkit and Safari) we stub out Object.freeze
using an identity function.  Test to see whether we have a functional
Object.freeze before using it in our test suite.


zloirock:
Thanks! I also already fixed it, but not committed :)

PR #60, opened 2015-04-24
Use our Promise implementation if native Promises don't subclass properly

cscott:
This fixes the subclass problems with Promise on Firefox.

A different alternative is to try to patch `Promise.prototype.{then,catch}` instead of overriding the entire Promise implementation... but then we might want to expand our subclass test coverage to double-check that `Promise.all`, `Promise.race`, etc don't have any subclass issues.

### DIFF
--- a/modules/es6.promise.js
+++ b/modules/es6.promise.js
@@ -5,6 +5,7 @@ var $       = require('./$')
   , $def    = require('./$.def')
   , assert  = require('./$.assert')
   , forOf   = require('./$.for-of')
+  , setProto = require('./$.set-proto')
   , SPECIES = require('./$.wks')('species')
   , RECORD  = require('./$.uid').safe('record')
   , PROMISE = 'Promise'
@@ -15,6 +16,7 @@ var $       = require('./$')
   , Base    = P
   , isFunction     = $.isFunction
   , isObject       = $.isObject
+  , $create        = $.create
   , assertFunction = assert.fn
   , assertObject   = assert.obj
   , test;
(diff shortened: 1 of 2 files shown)


modules/es6.promise.js
```diff
@@ -107,7 +107,22 @@ function $resolve(value){
 }
 
 // constructor polyfill
-if(!(isFunction(P) && isFunction(P.resolve) && P.resolve(test = new P(function(){})) == test)){
+var workingPromise = isFunction(P) && isFunction(P.resolve) && P.resolve(test = new P(function(){})) == test;
```
zloirock: long string

modules/es6.promise.js
```diff
@@ -107,7 +107,22 @@ function $resolve(value){
...
+// Firefox ~33 had broken subclass support, test that.
+if (workingPromise) {
+  var P2 = function(x) {
+    var self = new P(x);
+    Object.setPrototypeOf(self, P2.prototype);
```
zloirock: In modules insted of `Object.setPrototypeOf` should use `require('./$.set-proto').set` and in browsers w/o `Object.setPrototypeOf` or `__proto__` it's missing.

cscott:
Wrapped the test in a try/catch just to protect against weird cases where setPrototypeOf might throw on a native object, in addition to the standard missing/broken possibilities.

modules/es6.promise.js
```diff
@@ -107,7 +108,27 @@ function $resolve(value){
...
+// Firefox ~33 had broken subclass support, test that.
+if (workingPromise){
+  try { // protect against bad/buggy Object.setPrototype
+    var P2 = function(x) {
+      var self = new P(x);
+      setProto.set(self, P2.prototype);
+      return self;
+    };
+    setProto.set(P2, P);
+    P2.prototype = Object.create(P.prototype);
```
zloirock: Better use `$.create` instead of `Object.create`. Old browsers with `__proto__` and alternative `Promise` polyfill case.

modules/es6.promise.js
```diff
@@ -107,7 +108,27 @@ function $resolve(value){
...
+    P2.prototype.constructor = P2;
+    if (!(P2.resolve(5).then(function(){}) instanceof P2)) {
+      throw Error('broken subclass support');
+    }
```
zloirock: I think, `workingPromise = false`  better `throw`.

cscott:
Fixed that; also realized that I hadn't patched `library/modules`, so did that too.

zloirock:
> I hadn't patched `library/modules`

Files from `library/modules` copied at build step from `modules`, possible, but is not necessary.

PR #61, opened 2015-04-29
[WIP] ES7 String padding

romeovs:
implements the [string padding spec](http://wiki.ecmascript.org/doku.php?id=strawman:string_padding) proposal for ES7.

Note that I added an issue on that proposal because I think it is flawed, and I'm hoping they'll fix it soon.
So the spec might change.


zloirock:
Need add `es7.string.lpad` and `es7.string.rpad` to: `shim.js`, `es7/index.js`, `es7/string.js`, `fn/string`, `build/build.ls`.

Need add tests.

modules/$.string-pad.js
```diff
@@ -0,0 +1,50 @@
+// http://wiki.ecmascript.org/doku.php?id=strawman:string_padding
+'use strict';
+var $def = require('./$.def')
+  , $ = require('./$')
+  , repeat = require('../fn/string/repeat');
+
+module.exports = function(left) {
+  return function $pad(that, minLength, fillChar) {
```
zloirock: Possible export this function.

modules/$.string-pad.js
```diff
@@ -0,0 +1,50 @@
+// http://wiki.ecmascript.org/doku.php?id=strawman:string_padding
+'use strict';
+var $def = require('./$.def')
+  , $ = require('./$')
+  , repeat = require('../fn/string/repeat');
```
zloirock: Not sure about export from `fn` path, I'll fix it later.

modules/$.string-pad.js
```diff
@@ -0,0 +1,50 @@
...
+
+module.exports = function(left) {
+  return function $pad(that, minLength, fillChar) {
+    // 2. Let S be ToString(O).
+    var S = that.toString();
```
zloirock: `ToString` !== `.toString()`, use simple `String`.

modules/$.string-pad.js
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    // 5. Let fillLen be the number of characters in S minus intMinLength.
+    var fillLen = intMinLength - S.length;
+
+    if ( fillLen < 0 || fillLen === Infinity ) {
+      // 6. If fillLen < 0, then throw a RangeError exception.
+      // 7. If fillLen is +∞, then throw a RangeError exception.
+      throw new RangeError('Cannot satisfy string length ' +
+                            minLength + ' for string ' + JSON.stringify(S));
```
zloirock: Why `JSON.stringify`? It's already string.

modules/$.string-pad.js
romeovs: you mean just the `$pad`, with an extra parameter to denote left or right padding?

modules/$.string-pad.js
romeovs: I just did that to escape the string and add the correct quotes.

modules/$.string-pad.js
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    // 8. Let sFillStr be the string represented by fillStr.
+    // 9. If sFillStr is undefined, let sFillStr be a space character.
+    var sFillStr = fillChar && fillChar.toString ? fillChar.toString() : ' ';
```
zloirock: Use `=== undefined` and `String`.

modules/$.string-pad.js
zloirock: I don't think escaping required.

modules/$.string-pad.js
zloirock: Yep. Although not essential.

modules/$.string-pad.js
romeovs: It's just to nice the error message. Eg it looks like:
```
Cannot satisfy string length 10 for string "foobar"
```
instead of
```
Cannot satisfy string length 10 for string foobar
```

zloirock: `JSON.stringify` missing in IE7-, `core-js` not adds it for many reason.

zloirock:
Thanks!

zloirock:
сс @KevinGrandon @rwaldron guys, can you specify what result should be by [this proposal](http://wiki.ecmascript.org/doku.php?id=strawman:string_padding) on:
```js
'abc'.lpad(4, 'de')  // 'dabc' or 'eabc'?
'abc'.rpad(10, '𠮷') // 32bit chars and length
```

KevinGrandon:
I suppose the proposal needs to be updated to make this more clear. I'm curious what Rick and Dave say, but my assumption would be to fill from the "edge" outward, until the length is met, truncating where necessary. If possible, here is how I think this should behave:

> 'abc'.lpad(4, 'de') // 'dabc' or 'eabc'?
eabc

> 'abc'.lpad(6, 'de')
edeabc

> 'abc'.rpad(10, '𠮷') // 32bit chars and length
abc𠮷𠮷𠮷�
"�" being the first half of "𠮷" (\uD842)

romeovs:
@KevinGrandon Did you also see my issue on the spec I raised at your repo?
about not throwing a range error when the minimal length is already met.

KevinGrandon:
@romeovs Thanks, will respond there.

zloirock:
@KevinGrandon clearly, thanks.

Naddiseo:
:( I was literally in the [middle of implementing this](https://github.com/Naddiseo/core-js/tree/wip/string_padding). I was waiting back on the [bug I reported](https://bugs.ecmascript.org/show_bug.cgi?id=4332) in the spec, which as you mention, is incorrect.

Naddiseo:
@romeovs, one thing missing from your implementation is to check if `intMinLength` is less than 0. I suggested on the ecmascript bug tracker than it should throw a `RangeError`.

PR #66, opened 2015-05-11
Fix Math.hypot normalization issues

benelsen:
Math.hypot will now correctly use the largest (absolute) value for normalization. 
This also fixes a bug where hypot would return a negative result if the first argument is negative.


zloirock:
Thanks!

PR #81, opened 2015-06-16
Do not escape `/`

benjamingr:
Hey, updating the polyfill based on change in https://github.com/benjamingr/RexExp.escape, based on usage data escaping it isn't required (see the data folder for more motivation for this decision).

### DIFF
--- a/modules/es7.regexp.escape.js
+++ b/modules/es7.regexp.escape.js
@@ -1,5 +1,5 @@
 // https://github.com/benjamingr/RexExp.escape
 var $def = require('./$.def');
 $def($def.S, 'RegExp', {
-  escape: require('./$.replacer')(/[\/\\^$*+?.()|[\]{}]/g, '\\$&', true)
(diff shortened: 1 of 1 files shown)


zloirock:
This behavior was based on your `polyfill.js`, https://github.com/benjamingr/RexExp.escape/issues/12#issuecomment-111794784 , initially I thought you forgot add it to proposal text. Ok, I'll merge it, but my position - better escape it for code generation.

benjamingr:
@zloirock yes, this changed in the polyfill.js file too. If you have a compelling case for escaping it or can show me a code base in the wild that does this I'm all ears for this. This of course equally goes for any other character that might need escaping.

I've looked at over 100 thousand code bases in the wild (repos) and found less than 10 usages (the data and scripts used to generate it are open in the repo). People escaping with a `.replace` or a shim and then using the regexp constructor was a lot more common.

PR #88, opened 2015-07-03
Pubish individual modules to NPM

JakeChampion:
Being able to require in individual polyfills from NPM would be a great way for user's to consume parts of this library. When it comes to code that is going to live in a browser, having each module imported individually should help decrease the overall bundled size when using browserify/webpack.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@ node_modules/
 *.log
 *.bak
 *.swp
-
+client/modules
 Thumbs.db
 
 Desktop.ini
--- a/build/Gruntfile.ls
+++ b/build/Gruntfile.ls
@@ -45,24 +45,58 @@ module.exports = (grunt)->
         singleRun: true
         browsers: ['PhantomJS']
   grunt.registerTask \modularize ->
-    versions = <[es6 es7]>
-    versions.forEach((v) -> grunt.file.mkdir(path.resolve(\client, \modules, v)))
-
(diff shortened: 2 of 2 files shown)


zloirock:
Not all is so simple.
- Some modules interact through internal modules, you will have some problems.
- Total size single ES6 and ES7 modules in your proposal 800kb (!!!), shim.min.js less than 40kb.

JakeChampion:
When minified, all the es6 modules comes in at 272kb and 96kb gzipped.

Could you clarify the first bullet point? Do you mean only size problems?

zloirock:
> Could you clarify the first bullet point?

As I say, modules interact via internal common modules and it will cause some problems. In current version not many / serious, but they will be - fake `arguments` iterator will not work, some problems with symbols, `Function#toString` will be wrapped dozens of times, maybe something else.

> When minified, all the es6 modules comes in at 272kb and 96kb gzipped.

`build:es6,es7 uglify` -> 34kb, 12kb gzipped, 8x difference with your proposal, not serious.

jimmywarting:
I think Aurelia is struggling with what @zloirock is mentioning... A really cool project but hate to load a 600kb lib and use a preloader to wait for a Aurelia app to initiate.

JakeChampion:
@jimmywarting You can use other alternatives to core-js such as the es6-shim or polyfill-service

zloirock:
@JakeChampion with Aurelia? Definitely not.

PR #91, opened 2015-07-10
Get correct global object in Firefox extensions

### LINKED ISSUE #86: Unable to execute core-js in firefox's content_script environment
My team and I are using core-js via the babel/polyfill to develop a (cross) brower extension. 

During today's debugging session I found out that importing/requireing the polyfill or core-js causes SEGFAULTs in Firefox. Don't get me wrong I'm not gonna blame corejs for segfaults in FF :heart: 

So I went all the way down to the `modules/$.js` to find out that corejs references the `self` variable. 

https://github.com/zloirock/core-js/blob/master/modules/%24.js#L2

Obviously that's not a problem in most cases but for some reason the FF team thought it would be a great idea to use `self` as a namespace within an extensions content_script environment. 

https://developer.mozilla.org/en-US/Add-ons/SDK/Guides/Content_Scripts#Accessing_port_in_the_content_script

Honestly I didn't read your entire code base to find out why you're checking for `self`. With this issue I just want to rise a discussion if we somehow can fix this problem. 

Apart form this: Thanks for this great project :+1:

iamolivinius:
Fixes #86

Line 2-31 is taken from https://github.com/lodash/lodash/blob/3.10.0/lodash.js#L251-274

For more information see: https://github.com/zloirock/core-js/issues/86#issuecomment-120399476

### DIFF
--- a/modules/$.js
+++ b/modules/$.js
@@ -1,5 +1,5 @@
 'use strict';
-var global = typeof self == 'object' && self.Math == Math ? self : Function('return this')() 
+var global = typeof self == 'object' && self.Math == Math ? self : Function('return this')()
   , core   = {}
   , defineProperty = Object.defineProperty
   , hasOwnProperty = {}.hasOwnProperty
--- a/modules/$.js
+++ b/modules/$.js
@@ -1,35 +1,5 @@
 'use strict';
-var objectTypes = {
-  'function': true,
-  'object': true
-};
-
-/** Detect free variable `exports`. */
-var freeExports = objectTypes[typeof exports] && exports && !exports.nodeType && exports;
-
-/** Detect free variable `module`. */
-var freeModule = objectTypes[typeof module] && module && !module.nodeType && module;
-
-/** Detect free variable `global` from Node.js. */
-var freeGlobal = freeExports && freeModule && typeof global == 'object' && global && global.Object
-                 && global;
-
-/** Detect free variable `self`. */
-var freeSelf = objectTypes[typeof self] && self && self.Object && self;
-
-/** Detect free variable `window`. */
-var freeWindow = objectTypes[typeof window] && window && window.Object && window;
-
-/**
- * Used as a reference to the global object.
- *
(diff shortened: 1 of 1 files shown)


zloirock:
* Too many dead code.
* This code breaks CSP - `Function('return this')()` will be called in any case.
* You tried [this code](https://github.com/zloirock/core-js/issues/86#issuecomment-116287121)? By idea it should work.

iamolivinius:
I tried your code but it won't run in `bootstraped extensions`. Within bootstraped extensions `self === window` is true

zloirock:
Possible replace
```js
var sandbox = Function('return this')();
var global = root === sandbox ? root : sandbox
```
to
```js
var global = root === sandbox ? root : Function('return this')();
```
and it should work correct, but it looks too cumbersome.

Do you know any alternative way of detection FF extensions environment?

zloirock:
`var global = typeof self == 'object' && !self.port && self.Math == Math ? self : Function('return this')()`
?

iamolivinius:
Unfortunately I'm not aware of another way right now. I pinged the Mozilla Jetpack profile on twitter with the same question but without any response yet.

zloirock:
@iamolivinius try [this](https://github.com/zloirock/core-js/pull/91#issuecomment-120405925).

iamolivinius:
Clicked a hundred times on your link - nothing happened. Then I realized that it referenced the comment above :laughing: 

Well, https://github.com/zloirock/core-js/pull/91#issuecomment-120405925 also does the trick :+1: 

Works in Bootstraped as well as in add-on SDK extensions.

zloirock:
Try it also w/o `this.port`
```js
var global = typeof self == 'object' && self.Math == Math ? self : Function('return this')() 
```
not sure how works this object, update PR to working example and I'll merge it :)

iamolivinius:
:clap: It's also working without `self.port`

zloirock:
Thanks. Will be available in 1.0.

Issue #100, opened 2015-08-22
$.global is inaccessible after changing domain on IE 8


imsun:
```javascript
var global = typeof self != 'undefined' && self.Math == Math ? self : Function('return this')();
module.exports = global;
if(typeof __g == 'number')__g = global; // eslint-disable-line no-undef
```

In `$.global.js`, `self` is used as `global`. But on IE 8 if you changed `document.domain`, the elder `self` will be refused to access.

```javascript
// i.e.
var global = require('./$.global');
document.domain = 'foo.bar'

// no log on IE 8
global.console.log('Accessible!')
```

What's worse is that the loader will cache `$.global` module, so if you changed domain after the first time you load it, it won't work any longer.

zloirock:
You can propose more correct way `global` detection?

imsun:
Maybe you could detect if the script is executing in a web worker, and use `window` if not.

PR #101, opened 2015-08-22
Replace `self` by `window`

imsun:
Solution for the problem in https://github.com/zloirock/core-js/issues/100.


zloirock:
See [this table](https://github.com/zloirock/core-js/issues/86#issuecomment-115759028). In will break some environments.

zloirock:
Possible add check `window` like `self` *before* `self`, but it requires additional tests.

Issue #111, opened 2015-09-24
Closure Compiler Compatibility with Safari 8 and IE9 Minifying Array.from


jacobangel:
We ran into an issue where Array.from was minifying incorrectly. 

[...]

PR #112, opened 2015-09-24
Make Array.from more Minifier friendly.

jacobangel:
Certain minifiers will reuse references, which breaks in Safari8 and IE<10

Addresses Issue  #111


zloirock:
Ok, I'll merge it if it makes your calm, but it's not `core-js` problem and, as I say, with this Closure Compiler "optimisation" possible problems in many other places.

PR #114, opened 2015-09-30
Fix typo for importScript -> importScripts

abucur:
This typo is causing Firefox to improperly schedule tasks in WebWorker environments as setImmediate is not available there.


zloirock:
Thanks, really typo and problem in workers in some old, but still actual, FF versions. I'll publish new version in near time.

PR #115, opened 2015-10-02
Port Yaku to core-js

ysmood:
All the tests are passed, expect one test `the subclass`, and it may be a bug of the tests itself?

I don't know why it doesn't pass, I created a "tests/subclass-promise.js", it works well, it may be a bug of Karma?

Don't merge it before I fixed the `subclass` issue.

### DIFF
--- a/modules/es6.promise.js
+++ b/modules/es6.promise.js
@@ -55,160 +55,37 @@ var useNative = function(){
   return works;
 }();
 
-// helpers
-var isPromise = function(it){
-  return isObject(it) && (useNative ? classof(it) == 'Promise' : RECORD in it);
-};
-var sameConstructor = function(a, b){
-  // library wrapper special case
-  if(LIBRARY && a === P && b === Wrapper)return true;
-  return same(a, b);
-};
-var getConstructor = function(C){
-  var S = anObject(C)[SPECIES];
-  return S != undefined ? S : C;
-};
-var isThenable = function(it){
-  var then;
-  return isObject(it) && typeof (then = it.then) == 'function' ? then : false;
-};
-var notify = function(record, isReject){
-  if(record.n)return;
-  record.n = true;
-  var chain = record.c;
-  asap(function(){
-    var value = record.v
-      , ok    = record.s == 1
-      , i     = 0;
-    var run = function(react){
-      var cb = ok ? react.ok : react.fail
-        , ret, then;
-      try {
-        if(cb){
-          if(!ok)record.h = true;
-          ret = cb === true ? value : cb(value);
-          if(ret === react.P){
-            react.rej(TypeError('Promise-chain cycle'));
-          } else if(then = isThenable(ret)){
-            then.call(ret, react.res, react.rej);
-          } else react.res(ret);
-        } else react.rej(value);
-      } catch(err){
-        react.rej(err);
-      }
-    };
-    while(chain.length > i)run(chain[i++]); // variable length - can't use forEach
-    chain.length = 0;
-    record.n = false;
-    if(isReject)setTimeout(function(){
-      var promise = record.p
-        , handler, console;
-      if(isUnhandled(promise)){
-        if(isNode){
-          process.emit('unhandledRejection', value, promise);
-        } else if(handler = global.onunhandledrejection){
-          handler({promise: promise, reason: value});
-        } else if((console = global.console) && console.error){
-          console.error('Unhandled promise rejection', value);
-        }
-      } record.a = undefined;
-    }, 1);
-  });
-};
-var isUnhandled = function(promise){
-  var record = promise[RECORD]
-    , chain  = record.a || record.c
-    , i      = 0
-    , react;
-  if(record.h)return false;
-  while(chain.length > i){
-    react = chain[i++];
-    if(react.fail || !isUnhandled(react.P))return false;
-  } return true;
-};
-var $reject = function(value){
-  var record = this;
-  if(record.d)return;
-  record.d = true;
-  record = record.r || record; // unwrap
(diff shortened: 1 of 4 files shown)


modules/es6.promise.js
```diff
@@ -55,195 +45,66 @@ var useNative = function(){
...
-      notify(record, false);
-    }
-  } catch(e){
-    $reject.call({r: record, d: false}, e); // wrap
-  }
-};
 
 // constructor polyfill
 if(!useNative){
+  var Yaku = require('yaku');
```
benjamingr: We'll need a "dumbed down" version of yaku that is minimal in file size since this is included in browsers, as well as before-after measurements on the file size. It also has to be 100% compatible with native promises and the current polyfill w.r.t task scheduling.

ysmood: Dumbed down is a bad idea for npm ecosystem. You are already using webpack, I don't think there's any significant disadvantage to use external dependencies.

I'm already using the core-js' native task scheduling module `$.microstask`

benjamingr: Why? The polyfill should be self contained inside core-js so people including the shim in browsers would need to import the least amount of code possible - and not the whole library.

Anything that increases the file size isn't worth it IMO as people using node can use an external library (and not the shim anyway) and people in browsers don't create tens of thousands of promises anyway, so it should be as small and as compliant as possible.

ysmood: Give me some time, I will show you that the size will be decreased for core-js.

ysmood:
I'm on a vacation, the next commit may be several days later, sorry about that. I will get the code more neat.

benjamingr:
I think I speak for @zloirock when I say that any contribution and participation you offer to core-js is appreciated. Take your time and no rush.

modules/es6.promise.js
zloirock: For some reasons `core-js` should be self-contained.

modules/es6.promise.js
```diff
@@ -55,195 +45,66 @@ var useNative = function(){
...
+  // TODO: use core-js [[toString]]
+  // I haven't read much of the architecture of core-js, temporarily hard-coded it.
+  Yaku.toString = function () { return '[native code]'; };
+  Yaku.prototype.then.toString = function () { return '[native code]'; };
+  Yaku.prototype['catch'].toString = function () { return '[native code]'; };
+  Yaku.resolve.toString = function () { return '[native code]'; };
+  Yaku.reject.toString = function () { return '[native code]'; };
+  Yaku.all.toString = function () { return '[native code]'; };
+  Yaku.race.toString = function () { return '[native code]'; };
+  // TODO END
```
zloirock: `/native code/.test(method)` in tests - testing behavior `Function#toString` with these methods.

modules/es6.promise.js
```diff
@@ -55,195 +45,66 @@ var useNative = function(){
...
-    }
-  });
 }
 
 // export
 $def($def.G + $def.W + $def.F * !useNative, {Promise: P});
 require('./$.tag')(P, PROMISE);
 species(P);
-species(Wrapper = require('./$.core')[PROMISE]);
+species(require('./$.core')[PROMISE]);
```
zloirock: W/o it `@@species` will not be added to wrapper of native constructor in `library` version.

zloirock:
> All the tests are passed, expect one test the subclass, and it may be a bug of the tests itself?

As minimum, you have removed subclassing logic from `.all` and `.race`.

modules/es6.promise.js
ysmood: I tried it, but it seems to not work. My code:

```js
$mix(P.prototype, Yaku.prototype)
```

Why it doesn't work?

modules/es6.promise.js
```diff
@@ -55,209 +42,17 @@ var useNative = function(){
...
-    // 25.4.5.1 Promise.prototype.catch(onRejected)
-    'catch': function(onRejected){
-      return this.then(undefined, onRejected);
-    }
-  });
+  P = require('yaku');
+  require('yaku/lib/globalizeUnhandledRejection')();
+  P.nextTick = asap;
+
+  $mix(P, P);
```
ysmood: @zloirock Why `$mix` doesn't work?
The `toString` still outputs wrong string.

modules/es6.promise.js
```diff
@@ -55,209 +42,17 @@ var useNative = function(){
...
+  $mix(P.prototype, P.prototype);
 }
 
 // export
 $def($def.G + $def.W + $def.F * !useNative, {Promise: P});
```
ysmood: The `$def` doesn't expose the Promise, I cannot use `core.Promise` to get the Promise object, why?

modules/es6.promise.js
zloirock: Because target and source are the same. It saves result `toString` of a previous method if it's present.

modules/es6.promise.js
zloirock: Emm?

ysmood: Not in the browser, I mean node. By the way, is there a way to watch the test, run `npm test` again and again is boring. Or how you test the project when developing it, any tutorial or documentation?

zloirock: The same.
default
More convenient way for me - build browser bundles (`npm run grunt`), build  tests (`npm run grunt livescript`) and run tests in the browser (`tests`, `library` and other in the `tests` path). Or just use `karma` for automatize this process. Need to write docs.

zloirock:
3.5 month without activity - closed.

ysmood:
Finally I made it #170

PR #131, opened 2015-11-17
Use SVG instead of PNG for Travis CI badge

lencioni:
This will look better on a wider range of displays.


zloirock:
Ok, thanks :)

PR #132, opened 2015-11-17
Update `padLeft`/`padRight` to be `padStart`/`padEnd`,

Naddiseo:
... per TC39 meeting notes

Issue #137, opened 2015-11-24
[WIP] 2.0


zloirock:
- [x] Add implementations and fixes `ArrayBuffer`, `DataView`, all typed arrays constructors with all static and prototype methods. Really hard point. #90
- [x] Improve collections polyfills - `O(1)` and prevent possible leaking with frozen keys, correct observable state for object keys, #134
- [x] Add `{keys, values, entries, @@iterator}` methods to all DOM collections which should have [iterable interface](https://heycam.github.io/webidl/#idl-iterable) or should be [inherited from `Array`](https://heycam.github.io/webidl/#LegacyArrayClass)
- [x] Add basic support [`rejectionHandled` event](https://nodejs.org/api/process.html#process_event_rejectionhandled) / [`onrejectionhandled` handler](https://developer.mozilla.org/en-US/docs/Web/API/Window/onrejectionhandled) to the `Promise` polyfill
- [x] Rename `String#{padLeft, padRight}` -> `String#{padStart, padEnd}`, [proposal](https://github.com/tc39/proposal-string-pad-start-end), [November TC39 meeting](https://github.com/rwaldron/tc39-notes/tree/master/es7/2015-11/nov-17.md#conclusionresolution-2) (they want to rename it on each meeting?O_o), #132
- [x] Add `String#{trimStart, trimEnd}` as aliases for `String#{trimLeft, trimRight}`, [proposal](https://github.com/sebmarkbage/ecmascript-string-left-right-trim), [November TC39 meeting](https://github.com/rwaldron/tc39-notes/tree/master/es7/2015-11/nov-17.md#conclusionresolution-2)
- [x] Add `System.global`, [proposal](https://github.com/tc39/proposal-global), [November TC39 meeting](https://github.com/rwaldron/tc39-notes/tree/master/es7/2015-11/nov-19.md#systemglobal-jhd)
- [x] Add `Error.isError`, [proposal](https://github.com/ljharb/proposal-is-error), [November TC39 meeting](https://github.com/rwaldron/tc39-notes/tree/master/es7/2015-11/nov-19.md#jhd-erroriserror)
- [x] Remove usage `@@species` from `Promise.{all, race}`, [November TC39 meeting](https://github.com/rwaldron/tc39-notes/tree/master/es7/2015-11/nov-18.md#conclusionresolution-2)
- [x] Check changes by [meeting notes](https://github.com/rwaldron/tc39-notes/tree/master/es7/2015-11)
[...]

PR #142, opened 2015-12-11
Update README.md

yukulele:
https://github.com/zloirock/core-js/pull/132


zloirock:
Thanks for PR, but I can't merge it now. It was breaking change and it will be available in near days in `2.0.x` with many other changes, #137. Current `README` contains API version marked as `latest` - `1.2.x`, where those methods available by old names.

PR #160, opened 2016-01-14
fixed ie8 object assign 'Member not found' error.

cchamberlain:
I started receiving the following error after implementing Object.assign with an empty first object:

```
Object.assign({}, ...)
```

Image of error I get in modern ie XP VM on Windows 10 host:

image

Appears IE is throwing an error trying to assign:
```
T[key] = S[key];
```

But a check fixes the issue:
```
var sValue = S[key];
if(sValue)
  T[key] = sValue;
```

I published to npm to test as a separate package at core-js-ie8 and confirmed it fixed my issue.


zloirock:
Thanks for the PR, but I can't reproduce it. This PR also will break `Object.assign` in some cases. Can you show me a reproducible example?

cchamberlain:
Thanks.  I was using the shim via babel polyfill.  I'll put something together.

cchamberlain:
Looks like this was an error on my end that wasn't apparent until I built the non-shim version of my project.  I was attempting to use `Object.assign(DOMElement, { id: '...', innerHTML: '...', style: { ... } })` and looks like it was choking when trying to reassign the elements style property (for good reason).

I got the code refactored and fixed in chrome, have not tested IE8 yet but think its probably good now.  Going to close the issue.  If I do end up getting an IE8 bug I'll attempt to fix the branch and reopen.

Thanks!

PR #165, opened 2016-01-27
fix IE11 Script Access Denied error.

doctyper:
This pull request fixes an issue that surfaced when we used core-js on a page containing iframes in IE11. Attempting to access properties on the iframe threw `Script Access Denied` errors.

### DIFF
--- a/modules/_classof.js
+++ b/modules/_classof.js
@@ -16,9 +16,11 @@ module.exports = function(it){
   var O = Object(it);
 
   // @@toStringTag case
-  if (TAG in O && typeof O[TAG] === 'string') {
-    return O[TAG];
-  }
+  try {
+    if (typeof O[TAG] === 'string') {
+      return O[TAG];
+    }
+  } catch (e) {}
 
   var B = cof(O);
 
--- a/modules/_classof.js
+++ b/modules/_classof.js
@@ -5,12 +5,32 @@ var cof = require('./_cof')
   , ARG = cof(function(){ return arguments; }()) == 'Arguments';
 
 module.exports = function(it){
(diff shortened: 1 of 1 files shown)


modules/_classof.js
```diff
@@ -5,12 +5,32 @@ var cof = require('./_cof')
...
+  }
+
+  if (it === null) {
+    return 'Null';
+  }
+
+  var O = Object(it);
+
+  // @@toStringTag case
+  if (TAG in O && typeof O[TAG] === 'string') {
```
doctyper: The fix here was to add the `TAG in O` check.

zloirock:
Thanks for the PR! Can you show a reproducible example for clarifying the problem? Some fixes for IE11 `Script Access Denied` already added.

zloirock:
@doctyper so what about a reproducible example?

doctyper:
@zloirock It's very difficult to abstract this into a reproducible example.

We use a React stack using Babel to transpile our code. Babel uses core-js to shim functionality, and that's where the bug surfaces for us.

I've tried to reproduce the error using just core-js but the fact that the bug only occurs on IE11/IE Edge makes this a very tedious process.

I do believe this is in some way similar to the fallbacks you added (I used them to track down this bug as well).

zloirock:
The main problem here - `({get [Symbol.toStringTag](){ throw 42 }}) + ''` should throw error by the spec, but with `try / catch` it will not work. And performance issue. I also can't reproduce it. Need to think what we can do.

zloirock:
Still can't reproduce it, but OK, added this fallback.

PR #168, opened 2016-01-31
Better Promise unhandled rejection logging

jsdf:
The current logging of unhandled Promise rejections in the browser is not very helpful, as it does not print a stack trace with line numbers.

This PR prints a stack trace when available, which in most browsers results in clickable line numbers being displayed, linking back to the original error source. This provides a better developer experience when tracking down the root cause of unhandled promise rejections.

### DIFF
--- a/modules/es6.promise.js
+++ b/modules/es6.promise.js
@@ -135,7 +135,7 @@ var onUnhandled = function(promise){
       } else if(handler = global.onunhandledrejection){
         handler({promise: promise, reason: value});
       } else if((console = global.console) && console.error){
-        console.error('Unhandled promise rejection', value);
+        console.error('Unhandled promise rejection', isObject(value) && value.stack || value);
       } promise._h = 2;
     } promise._a = undefined;
   });


loganfsmyth:
I'm curious what browser you are using that was having trouble. Chrome for instance will print the error with `(...)` after the message and clicking on it will show the stacktrace.

zloirock:
It does not look good for me:

- IE `Error` hasn't non-standard `stack` property.
- In Chrome it looks good - `stack` contains a message, but in FF - not. Also, in FF current stack trace output looks more pretty then proposed.
- If promise will be rejected with a non-error object - also will be used `stack` property.

As @loganfsmyth wrote - you can use `(...)` in Chrome.

That confuses some users, so feel free propose improved solution without problems, mentioned above.

jsdf:
Fair enough, after looking at some more browsers it seems quite difficult to come up with a good inconsistent. The browsers I had tested it in were Chrome 47, IE10, and FF. I had another look at FF I see what you're saying about the output being worse.

It's true that IE<10 doesn't define the stack property, but as the code is written it should just fall back to printing `value` in that case, though it could be made more explicit.

In IE10 printing `value` only prints the error description, while `value.stack` prints the full stack trace. In IE11 and Edge the output of printing `value` displays an object dump of the error object's properties, while printing `value.stack` again prints the stack trace (though the clickable links don't do anything useful at this stage). Recent versions of Safari have similar behaviour to IE>=10.

It seems that printing the stack property (when it is present) is desirable in all browsers except Firefox. I guess the tough part is how to detect when to use which method of printing the error.  One possibility is to check if the value of the `error.message` property exists within `error.stack`, and if so print the stack, otherwise print the error. What do you think about that?

zloirock:
I played with it and removed those main problems, but still  not sure it should be added - it's not very useful without long stack trace support.

zloirock:
In this case:
```js
if(HAS_STACK && classof(value) == 'Error'){
  if(STACK_VALUE)console.error(UNHANDLED_MESSAGE, value.stack);
  else console.error(UNHANDLED_MESSAGE, value.message, '\n' + value.stack);
} else console.error(UNHANDLED_MESSAGE, value);
```
Safari looks ugly. Any ideas?

zloirock:
Closed because of inactivity.

Issue #159, opened 2016-01-09
es6.promise.js feature testing causes console error in Chrome Canary v49.0.2615.0?


coopy:
```
Uncaught (in promise) TypeError: promiseCapability.[[Resolve]] is not a function
    at Object.then (native)
```

I and some others are seeing this error pop up in React apps running in Chrome Canary. The theory is that this error is emanating from `babel-polyfill`, which is depending on `core-js`. See discussion at https://code.google.com/p/chromium/issues/detail?id=575314. Chromium seems to have enabled more strict checks.

Repro:

1. `git clone git@github.com:Ognian/test.git`
2. `cd test && npm install && npm start`
3. Visit `http://localhost:3002` in Chrome Canary `v49.0.2615.0`, with developer tools console open

Expected: no error
Actual: Error is displayed:

```
Uncaught (in promise) TypeError: promiseCapability.[[Resolve]] is not a function
    at Object.then (native)
```

zloirock:
Thanks for the report. In Chrome 49, `core-js` currently uses native promises - looks like the bug is here. I will test it.

zloirock:
```js
var test = new Promise(function(){});
test.constructor = Object;
Promise.resolve(test);
```
is buggy in Chrome 49 native promises. And it creates serious problems for testing subclassing.

littledan:
What behavior are you observing, and what do you expect?

zloirock:
@littledan subclassing. By the spec:
```js
var test = new Promise(function(){});
Promise.resolve(test) === test; // should be `true`

var test = new Promise(function(){});
test.constructor = function(){};
Promise.resolve(test) === test; // should be `false`
```
But in Chrome 49 the second example creates rejected promise - bug. Why? Looks like `.then` uses `.constructor` instead of `@@species`.

littledan:
It's true that we don't yet implement @@species (sorry, working on it!), but I don't think that's the cause here. For me, both of those tests pass. Which version of Chrome are you running? I tested the version of V8 that's in 49.0.2617.

zloirock:
> It's true that we don't yet implement `@@species`

The problem not in `@@species` implementation. `SpeciesConstructor` **never** returns just `.constructor` - it's buggy behavior.

> but I don't think that's the cause here

It will not be caused here only with other serious bugs. I did not seek it in V8 source code - I just opened required method for confirming my guess.

[...]

PR #170, opened 2016-02-15
Port Yaku to core-js

ysmood:
Continue after the #115. Sorry for so long. This time everything goes well. Only one file changed (92162b40f4cc6ee38a575ba5f28a82e90b4cf912), all the tests should pass.

When I debug the project, most of the unit test time is taken by the promise, it's unbearable for new contributors. Pulling the core code of promise out of the project will make the project more maintainable.

You don't have to run the `promises-aplus-tests` any more. Because as a separate project, yaku has already maintained the `promises-aplus-tests` on itself. So that you can focus on other parts of the project. Besides, I tested core-js against the [promises-es6-tests](https://github.com/promises-es6/promises-es6), it didn't pass all the tests. Yaku did pass them all.

With Yaku, the performance will be better, the error handling will be less painful (such as long stack trace support), the gzipped size of the core.min.js won't change (27KB).

Here's the comparison:

| name | unit tests | 1ms async task | optional helpers | helpers | min js |
| --- | --- | --- | --- | --- | --- |
| [yaku](https://github.com/ysmood/yaku)@0.11.6 | ✓ | 330ms / 106MB | ✓ | 29 | 3.8KB |
| [bluebird](https://github.com/petkaantonov/bluebird)@3.3.1 | x (18 failing) | 265ms / 88MB | partial | 100 | 52.2KB |
| [es6-promise](https://github.com/jakearchibald/es6-promise)@3.1.2 | x (52 failing) | 426ms / 113MB | x | 10 | 6.3KB |
| [native](http://people.mozilla.org/~jorendorff/es6-draft.html#sec-promise-objects)@0.12.4 | x ( 4 failing) | 590ms / 173MB | x | 13 | 0KB |
| [core-js](https://github.com/zloirock/core-js)@2.1.0 | x ( 4 failing) | 838ms / 198MB | x | 11 | 13.9KB |
| [es6-shim](https://github.com/paulmillr/es6-shim)@0.34.4 | ✓ | 950ms / 145MB | x | 12 | 130.8KB |
| [q](https://github.com/kriskowal/q)@1.4.1 | x (47 failing) | 1599ms / 425MB | x | 74 | 15.4KB |

### DIFF
--- a/modules/es6.promise.js
+++ b/modules/es6.promise.js
@@ -1,26 +1,12 @@
 'use strict';
-var LIBRARY            = require('./_library')
-  , global             = require('./_global')
-  , ctx                = require('./_ctx')
-  , classof            = require('./_classof')
+var global             = require('./_global')
   , $export            = require('./_export')
-  , isObject           = require('./_is-object')
-  , anObject           = require('./_an-object')
-  , aFunction          = require('./_a-function')
-  , anInstance         = require('./_an-instance')
-  , forOf              = require('./_for-of')
-  , setProto           = require('./_set-proto').set
-  , speciesConstructor = require('./_species-constructor')
-  , task               = require('./_task').set
-  , microtask          = require('./_microtask')
+  , redefine           = require('./_redefine')
+  , wks                = require('./_wks')
   , PROMISE            = 'Promise'
-  , TypeError          = global.TypeError
-  , process            = global.process
   , $Promise           = global[PROMISE]
-  , process            = global.process
-  , isNode             = classof(process) == 'process'
   , empty              = function(){ /* empty */ }
-  , Internal, GenericPromiseCapability, Wrapper;
+  , Yaku               = require('./yaku');
 
 var USE_NATIVE = !!function(){
   try {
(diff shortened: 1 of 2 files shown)


zloirock:
Thanks for the PR, but I'm already not sure about adding dependencies to `core-js`. Some conceptions which I wanna use in the future makes it problematic. Need to think about it.

Without any tests I see some problems:

- Subclassing should use [`@@species` pattern](https://tc39.github.io/ecma262/#sec-speciesconstructor), but uses `this.constructor`, [it's completely incorrect](https://github.com/zloirock/core-js/issues/159).
- [Iteration closing](https://tc39.github.io/ecma262/#sec-iteratorclose) in `Promise.all` and `Promise.race` and some other problems with iterators.
- Static methods have separate feature detection and polyfilling. For example, `.resolve` should be always replaced in `library` version for some reasons.
- In environments without `Symbol`, well-known symbols `@@iterator` and `@@species` will not work in `library` version or if `es6.symbol` module will be required after `es6.promise`.

> I tested core-js against the promises-es6-tests, it didn't pass all the tests.

I know about it, it's a part of my not public `Promise` test case. `core-js` didn't pass 2 asserts because of 1 minor bug with the order of resolving. It's a minor problem, I was going to fix it later. Other 2 asserts can fail only with full `core-js` version because of [`Number` iterator](https://github.com/zloirock/core-js#number-iterator) - use `shim` version for this test case.

What do you think about replacing `core-js` `Promise` core to modified `Yaku` `Promise` core instead of adding it as a dependency?

ysmood:
> - Subclassing in .then should use @@species pattern, but uses this.constructor, it's completely incorrect.

Fixed 

> - Iteration closing in Promise.all and Promise.race and some other problems with iterators.
> - Static methods have separate feature detection and polyfilling. For example, .resolve should be always replaced in library version for some reasons.
> - In environments without Symbol, well-known symbols @@iterator and @@species will not work in library version or if es6.symbol module will be required after es6.promise

Fixed 

> - Support rejectionHandled event.

Added

zloirock:
Something like that makes sense. I will explore it when I will have spare time.

ysmood:
No harry, please take your time. I will add more test cases, like these:

https://github.com/ysmood/yaku/blob/master/test/basic.js

https://github.com/ysmood/yaku/blob/master/test/basic.js

zloirock:
Closed because of inactivity.

PR #174, opened 2016-02-25
Fixes #173. Prevents RedExp.exec() crash in old Qt Script engine (WebKit) when using Symbol()

### LINKED ISSUE #173: `Symbol()` polyfill causes RegExp.exec() to crash in old WebKit.
I'm adding the `core-js` polyfill to Qt's scripting environment, [Qt Script](https://doc.qt.io/qt-5/qtscript-index.html). `Qt Script` is a [WebKit/JavaScriptCore engine](https://github.com/qtproject/qtscript/tree/dev/src/3rdparty/javascriptcore/JavaScriptCore) that dates back to 2011.

I've created the polyfill/shim with this command:
```
npm run grunt build:es5,es6,es7,js,web -- --blacklist=es6.date.to-iso-string,es6.object.assign,es6.weak-map,es7.reflect --path=dist/core-js-shim-polyfill
```

I have to blacklist a few modules as they cause issues/crashes:
  * `es6.date.to-iso-string`
  * `es6.object.assign`
  * `es6.weak-map`
  * `es7.reflect`

I'm trying to get `es6.object.assign` to work. If I do not include it in the blacklist, so it IS in the build, loading the polyfill crashes the `Qt Script` environment. I have tracked the crash down to this call to `RegExp.exec()`:
  * https://github.com/zloirock/core-js/blob/7a90ad4004a437bbbf96e29051904eedda3f7457/modules/es6.regexp.split.js#L18

I have found that after the `Object.assign()` module calls `Symbol()`, any call to `RegExp.exec()` will crash the environment. The call to `Symbol()` is located here:
  * https://github.com/zloirock/core-js/blob/7a90ad4004a437bbbf96e29051904eedda3f7457/modules/_object-assign.js#L14

Adding a simple `RegExp.exec()` call right after that line will crash the environment:
```
var crash = function () {
  var myPattern = /^([a-z0-9.+-]+:)/i;
  var rest = 'foo:bar';
  var myArray = myPattern.exec(rest);
  return myArray;
};
crash();
```

The `Qt Script` environment appears to break when `Object.assign()`'s call to `Symbol()` calls `Object.defineProperty()` which `setSymbolDesc` is a reference to here:
  * https://github.com/zloirock/core-js/blob/7a90ad4004a437bbbf96e29051904eedda3f7457/modules/es6.symbol.js#L55

After that call, I see a new `Symbol` all over the `prototype` of most objects (Array, Object, RegExp, etc.):
## `Object.prototype` before call to `Symbol()`:
object-before-call

## `Object.prototype` after call to `Symbol()`:
object-after-call

## `Regexp.prototype` before call to `Symbol()`:
regexp-before-call

## `Regexp.prototype` after call to `Symbol()`:
regexp-after-call

Calling `var foo = Symbol();` anywhere seems to cause this problem as well. I'm not sure what the issue here is. Should `Object.assign()`'s call to `Symbol()` pass it a parameter instead of `undefined`? There are no other calls to `Symbol()` with an `undefined` parameter in the full shim polyfill code.

I believe this `Symbol()_g.dmlijp036315rk9` on the `RegExp.prototype.__proto__` is causing the crash.

bendiy:
Symbol()'s call to Object.defineProperty() accessor descriptor should set a default 'get' property when adding a 'set' property. Prevents crash in old Qt Script engine (WebKit).

### DIFF
--- a/library/modules/es6.symbol.js
+++ b/library/modules/es6.symbol.js
@@ -54,10 +54,8 @@ var wrap = function(tag){
   sym._k = tag;
   DESCRIPTORS && setter && setSymbolDesc(ObjectProto, tag, {
     configurable: true,
-    set: function(value){
-      if(has(this, HIDDEN) && has(this[HIDDEN], tag))this[HIDDEN][tag] = false;
-      setSymbolDesc(this, tag, createDesc(1, value));
-    }
+    enumerable: false,
+    writable: true
   });
   return sym;
 };
--- a/modules/es6.symbol.js
+++ b/modules/es6.symbol.js
@@ -54,10 +54,8 @@ var wrap = function(tag){
   sym._k = tag;
   DESCRIPTORS && setter && setSymbolDesc(ObjectProto, tag, {
     configurable: true,
-    set: function(value){
-      if(has(this, HIDDEN) && has(this[HIDDEN], tag))this[HIDDEN][tag] = false;
-      setSymbolDesc(this, tag, createDesc(1, value));
-    }
+    enumerable: false,
+    writable: true
   });
   return sym;
 };


bendiy:
Closing. This doesn't fix the problem in #173.

PR #181, opened 2016-03-17
Fix Object.getOwnPropertyNames(window) replacement

AgentME:
There's an issue that would cause `Object.getOwnPropertyNames(window)` to return outdated results in some browsers. It would try to call the native `Object.getOwnPropertyNames` and then fall back, but there was a mistake in the try block so it would always fall back.

(This caused some really fun issues when mixing Babel code where some used the transform-runtime plugin and some didn't. Before I realized the cause of it was so simple, I started documenting this [example](https://github.com/AgentME/regenfail) because I imagined the fix would be much more involved.)


zloirock:
Thanks, looks like a bug after splitting internal modules in `2.1`.

PR #192, opened 2016-04-20
add `Symbol.asyncIterator`

HyeonuPark:
[Async iteration proposal](https://github.com/tc39/proposal-async-iteration) adds another popular symbol `Symbol.asyncIterator`, which is used for async for-of statement.

I'm not very familiar with sending PR to another repo, so please let me know if something is bad. :smile:

### DIFF
--- a/es7/symbol.js
+++ b/es7/symbol.js
@@ -1,0 +1,2 @@
+require('../modules/es7.symbol.async-iterator');
+module.exports = require('../modules/_core').Symbol;
--- a/modules/es7.symbol.async-iterator.js
+++ b/modules/es7.symbol.async-iterator.js
@@ -1,0 +1,10 @@
+'use strict';
+// Symbol.asyncIterator shim
+var core    = require('./_core')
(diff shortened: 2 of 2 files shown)


zloirock:
Hi, @HyeonuPark. That definitely should be added, thanks for the PR. But it's a little harder what do you think. For old engines `wks` helper returns internal unique identifier, it should be wrapped like in `es6.symbol` module. I think, I will add internal helper for that and add `Symbol.asyncIterator` and `Symbol.observable` manually in the next release. Or you can play with this helper.

HyeonuPark:
Thanks for your reply @zloirock! I understand this issue is a bit more complicated then i expected.

Can I ask you when are you planning to make that helper? I don't want to urge you, but it will be helpful to such people like me who want to use those features.

zloirock:
In near days when I will have some free time. I think this week.

zloirock:
[Added.](https://github.com/zloirock/core-js/releases/tag/v2.3.0)

PR #197, opened 2016-05-14
Close unclosed script tag

james-gibson:


loganfsmyth:
The ` + gt` already adds this on.

zloirock:
Yep, `+ gt` closes this tag. It will break HTML if it will be inlined to HTML document as one string.

Issue #186, opened 2016-03-30
es6.promise always pause on exception in Canary


szarouski:
Hi, I can't use pause on exception feature in Canary (51.0.2694.1 canary (64-bit)) because es6.promise causes "Iterator result undefined is not an object" type error. Is there a way to fix that?

![temp](https://cloud.githubusercontent.com/assets/911388/14160561/ff054848-f6aa-11e5-8ac1-3bed74a49646.gif)

Thank you.

zloirock:
Interesting. Looks like exceptions in promises, even with handlers, causes this problem:
```js
new Promise(_ => { throw 42 }).catch(e => console.log(e));
```
It's Chrome dev tools issue, not core-js. I can try to add a fallback for this example, but not sure it's possible.

PR #210, opened 2016-06-13
v1.x.x backport request: Add a fallback for Chrome dev tools strange behavior

loklaan:
This pr aims to backport an already implemented fix (see #186) to the 1.x.x version track, for a breaking change introduced by the current stable version of Chrome.

_There doesn't seem to be a 1.x.x version track for core-js, but I thought I'd try to this request anyway._ 😉 

This could be seen as a critical bugfix, since babel 5.x.x still uses the core-js in the `^1.0.0` range. I can imagine quite a few devs who are maintaining an app with babel 5.x.x would appreciate this backport.

Also, thanks for the original fix! You've made it easy to apply a temporary hotfix to my own app.

**Note: I don't intend this pr to actually be merged into `master`. Perhaps you could create a `1.x.x` branch, if you go ahead with this pr?**

### DIFF
--- a/library/modules/$.iter-detect.js
+++ b/library/modules/$.iter-detect.js
@@ -13,7 +13,7 @@ module.exports = function(exec, skipClosing){
   try {
     var arr  = [7]
       , iter = arr[ITERATOR]();
-    iter.next = function(){ safe = true; };
+    iter.next = function(){ return {done: safe = true}; };
     arr[ITERATOR] = function(){ return iter; };
     exec(arr);
   } catch(e){ /* empty */ }
--- a/modules/$.iter-detect.js
+++ b/modules/$.iter-detect.js
@@ -13,7 +13,7 @@ module.exports = function(exec, skipClosing){
   try {
     var arr  = [7]
       , iter = arr[ITERATOR]();
-    iter.next = function(){ safe = true; };
+    iter.next = function(){ return {done: safe = true}; };
     arr[ITERATOR] = function(){ return iter; };
     exec(arr);
   } catch(e){ /* empty */ }


loklaan:
Update on this PR (I am closing it).

This approach may still be useful to some people that currently have issues with devtools in Chrome, that have to work with core-js v1.x.x.

_But_, the issue with Babel 5.x.x that drove me to organise this PR doesn't actually exist... I was simply led astray by my own false conclusions after debugging the overall issue. The latest `babel-polyfill` is using an up-to-date `core-js`, which should resolve this devtools error.

Also, to those that are led here for similar reasons as mine were to create this PR, you should know that upgrading from babel 5 to babel 6 isn't _that terrible_. I recommend these [two](https://medium.com/@malyw/how-to-update-babel-5-x-6-x-d828c230ec53) [articles](https://medium.com/@kentcdodds/misunderstanding-es6-modules-upgrading-babel-tears-and-a-solution-ad2d5ab93ce0) for getting a quick grok on potential downfalls.

loganfsmyth:
If this is in fact an issue, it may still be worth doing, since `babel-runtime@5.x` is still on `core-js@1.x`, as is anyone using `babel-core/polyfill` from `babel@5.x`.

loklaan:
@loganfsmyth It wouldn't hurt to go ahead with it, yeah. (famous last words)

I'm happy to prove it is an issue. Do you have suggestions, or preferences, for me to provide proof?

dfreeman:
This definitely appears to be a problem for Babel 5 folks.

If I'm understanding the flow correctly, `core-js` would need a 1.x branch coming off of the `1.2.6` tag, and then the fix could be backported to that branch and released (as `1.2.7`?).

After that, there would need to be a fresh release of the `babel-*@5.x` packages that include `core-js`, since (I think) it gets baked into the larger polyfill artifact that's released to npm.

Issue #204, opened 2016-05-31
Restrictions for contents of script elements (_object-create.js)


epiloque:
In-lining the `core-js/library/modules/_object-create.js` in html `<script>` tag may theoretically backfire, `_object-create.js` has `<script>` tags in the source.

Should this be adressed ?

> [Restrictions for contents of script elements](https://www.w3.org/TR/html5/scripting-1.html#restrictions-for-contents-of-script-elements)

> The easiest and safest way to avoid the rather strange restrictions described in this section is to always escape "<!--" as "<\!--", "<script" as "<\script", and "</script" as "<\/script" when these sequences appear in literals in scripts (e.g. in strings, regular expressions, or comments), and to avoid writing code that uses such constructs in expressions. Doing so avoids the pitfalls that the restrictions in this section are prone to triggering: namely, that, for historical reasons, parsing of script blocks in HTML is a strange and exotic practice that acts unintuitively in the face of these sequences.

[_object-create.js#L23](https://github.com/zloirock/core-js/blob/e482646353b489e200a5ecccca6af5c01f0b4ef2/library/modules/_object-create.js#L23)

```js

  iframeDocument.open();
  iframeDocument.write('<script>document.F=Object</script' + gt);
  iframeDocument.close();

```

zloirock:
We had a problem with `</script>`, so now it's splitted, but not with `<script>`. Anyway, just in case, I'll fix this issue later.

epiloque:
Would you accept a PR ?

On 06/03, Denis Pushkarev wrote:
> We had a problem with `</script>`, so now it's splitted, but not with `<script>`. Anyway, just in case, I'll fix this issue later.
> 
> ---
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/zloirock/core-js/issues/204#issuecomment-223697463

-- 

Mark Milstein

@epiloque
+1-917-719-1790
http://bit.ly/pgp-key-mm

zloirock:
Sure.

PR #216, opened 2016-06-29
#204 Escaping script tags

funklos:
We experienced issues with inlining this code and parsing by facebook parser. So it's indeed safer to incorporate this.

### DIFF
--- a/modules/_object-create.js
+++ b/modules/_object-create.js
@@ -20,7 +20,7 @@ var createDict = function(){
   // html.removeChild(iframe);
   iframeDocument = iframe.contentWindow.document;
   iframeDocument.open();
-  iframeDocument.write('<\script>document.F=Object<\/script' + gt);
+  iframeDocument.write(lt + 'script' + gt + 'document.F=Object' + lt + '/script' + gt);
   iframeDocument.close();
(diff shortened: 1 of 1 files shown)


loganfsmyth:
Seems like it would be best to handle this like the `gt` does in the existing codebase. Relying on `\s` to convert back to `s` seems a little gross.

modules/_object-create.js
```diff
@@ -20,7 +20,7 @@ var createDict = function(){
   // html.removeChild(iframe);
   iframeDocument = iframe.contentWindow.document;
   iframeDocument.open();
-  iframeDocument.write('<\script>document.F=Object<\/script' + gt);
+  iframeDocument.write(lt + 'script' + gt + 'document.F=Object' + lt + '/script' + gt);
```
zloirock: `lt` is not defined.

funklos: lt is now defined

zloirock:
OK, thanks.

funklos:
When are you planning to release the next minor/patch version so we can update to get the fix?

zloirock:
@funklos after some days I'll publish patch release, after that I'll work on major release when I'll have time.

PR #218, opened 2016-07-10
Enable no-unused-vars and remove them

teppeis:
There are some unused vars and requires.
I recommend to enable `no-unused-vars`.
http://eslint.org/docs/rules/no-unused-vars


fn/date/to-string.js
```diff
@@ -1,5 +1,5 @@
-var core      = require('../../modules/es6.date.to-string')
-  , $toString = Date.prototype.toString;
+require('../../modules/es6.date.to-string')
```
teppeis: only side-effect

modules/es6.typed.array-buffer.js
```diff
@@ -6,7 +6,6 @@ var $export      = require('./_export')
   , toIndex      = require('./_to-index')
   , toLength     = require('./_to-length')
   , isObject     = require('./_is-object')
-  , TYPED_ARRAY  = require('./_wks')('typed_array')
```
teppeis: This doesn't have side effect, right?

zloirock:
LGTM, thanks.

Issue #236, opened 2016-09-03
es6 array iterator polyfill in_iter-define: "Safari has buggy iterators w/o `next`" - is the logic for this actually correct?


rubyruy:
I am referring to [this line from this file](https://github.com/zloirock/core-js/blob/3a98a54e6ffd9d616b82de0cf1edd666d9a219ad/modules/_iter-define.js#L12):

``` js
  , BUGGY          = !([].keys && 'next' in [].keys()) // Safari has buggy iterators w/o `next`
```

it is used when deciding to override `Array.prototype[Symbol.iterator]`, here:

``` js
 // Define iterator
  if((!LIBRARY || FORCED) && (BUGGY || VALUES_BUG || !proto[ITERATOR])){
    hide(proto, ITERATOR, $default);
  }
```

The problem is the value of `$default` simply grabs the existing native implementation which we just decided is buggy:

``` js
   , $native    = proto[ITERATOR] || proto[FF_ITERATOR] || DEFAULT && proto[DEFAULT]
    , $default   = $native || getMethod(DEFAULT)
```

It will use either `Array.prototype[Symbol.iterator]` or `Array.prototype.values`.

So either this logic is wrong, or this specific Safari bug _only_ affects the iterator returned by `Array.prototype.keys` (and not `.values` or `[Symbol.iterator]`) which seems unlikely. Why bother overwriting `hide(proto, ITERATOR, $default);` in that case?

I ran into this while working with a page inside [EAWebKit](https://github.com/xebecnan/EAWebkit). Of course this is not an officially supported engine, however, its JS engine definitely exhibits this exact problem (`[].keys()` returns an iterator without a `.next()` method - as does `.values()`). My guess is EA just so happened to fork off the same buggy version of WebKit as Safari did when this behavior was dealt with. 

core-js definitely fails to polyfill Array iterators in EAWebKit, so this leaves two possibilities:
1. Either the core.js logic for this fix is wrong and Safari patched the bug before anyone noticed, _or_
2. The bug core.js is trying to fix really did only affect the keys iterator, in which case EAWebKit either introduced the bug or forked a weird version of WebKit nobody has really looked at.

In the latter case we need to seriously re-evaluate EAWebKit and be on the lookout for all sorts of bugs unique to it, so I am very keen on finding out which it actually is.

If it's the former case I am happy to provide a PR.

rubyruy:
[...]

PR #237, opened 2016-09-03
default iterator should not use native implementation when buggy

rubyruy:
The fix for #236 (assuming the problem is valid)


zloirock:
Sorry that I missed it.

PR #238, opened 2016-09-05
Add Promise#finally proposal Stage 2 TC39 0716

clavecoder:
## [Promise#finally proposal Stage 2 from TC39 meeting on 2016-07](https://github.com/rwaldron/tc39-notes/blob/master/es7/2016-07/jul-28.md#10iiib-promiseprototypefinally).
- [proposal](https://github.com/rwaldron/tc39-notes/blob/master/es7/2016-07/jul-28.md#10iiib-promiseprototypefinally)
- [polyfill](https://github.com/tc39/proposal-promise-finally/blob/master/polyfill.js)

Used implementation of `String#padStart`, `String#trimLeft`, and `Promise#then` as examples. 
## Initial open items
- [X] Not able to use `finally` for the function name, perhaps because it is a reserved keyword. Temporarily used `_finally` for implementation and test (commented below).
- [X] Did not use polyfill's method of setting species. Coded to assume `core-js` Promise polyfill will set species, but, if I'm not mistaken, it's possible that `finally` may be `require`d without the Promise polyfill. Since `core-js` has its own method of setting species, what should be done?
- [ ] The failing test is precisely about Promise not having the right species (I think), but I'm not sure why, it shows up in the commonjs test for `delay`
  image
- [ ] The `finally` implementation is not in an internal file, but directly in the imported one. I am not supposing that it would be used by other functions in the library. Should I rather assume that it may be?

### DIFF
--- a/library/modules/es7.promise.finally.js
+++ b/library/modules/es7.promise.finally.js
@@ -1,0 +1,30 @@
+'use strict';
+// https://github.com/tc39/proposal-promise-finally
+var $export = require('./_export')
+  , core = require('./_core')
+  , global = require('./_global')
+  , speciesConstructor = require('./_species-constructor')
+  , $Promise = core.Promise || global.Promise;
+
+$export($export.P, 'Promise', {
+  'finally': function(onFinally){
+    var handler = typeof onFinally === 'function' ? onFinally : function(){};
+    var C = speciesConstructor(this, $Promise);
+    var newPromise = $Promise.prototype.then.call(
+      this, // throw if IsPromise(this) is not true
+      function(x) {
+        return new C(function(resolve) {
+          return resolve(handler());
+        })
+        .then(function(){ return x })
+      },
+      function(e) {
+        return new C(function(resolve) {
+          return resolve(handler());
+        })
+        .then(function(){ throw e; });
+      }
+    );
+    return newPromise;
+  }
+});
--- a/modules/es7.promise.finally.js
+++ b/modules/es7.promise.finally.js
@@ -1,0 +1,30 @@
+'use strict';
+// https://github.com/tc39/proposal-promise-finally
+var $export = require('./_export')
+  , core = require('./_core')
+  , global = require('./_global')
+  , speciesConstructor = require('./_species-constructor')
+  , $Promise = core.Promise || global.Promise;
+
+$export($export.P, 'Promise', {
+  'finally': function(onFinally){
+    var handler = typeof onFinally === 'function' ? onFinally : function(){};
+    var C = speciesConstructor(this, $Promise);
+    var newPromise = $Promise.prototype.then.call(
+      this, // throw if IsPromise(this) is not true
+      function(x) {
+        return new C(function(resolve) {
+          return resolve(handler());
+        })
+        .then(function(){ return x })
+      },
+      function(e) {
+        return new C(function(resolve) {
+          return resolve(handler());
+        })
+        .then(function(){ throw e; });
+      }
+    );
+    return newPromise;
+  }
+});
--- a/tests/tests/es7.promise.finally.ls
+++ b/tests/tests/es7.promise.finally.ls
@@ -1,0 +1,71 @@
+{module, test} = QUnit
+module \ES7
+
+test 'Promise#finally' (assert)!->
+  assert.isFunction Promise::finally
+  NATIVE and assert.arity Promise::finally, 2
+  assert.looksNative Promise::finally
+  assert.nonEnumerable Promise::, \finally
+  # subclassing, @@species pattern
+  promise = new Promise !-> it 42
+  promise@@ = FakePromise1 = !-> it ->, ->
+  FakePromise1[Symbol?species] = FakePromise2 = !-> it ->, ->
+  assert.ok promise.finally(->) instanceof FakePromise2, 'subclassing, @@species pattern'
+  # subclassing, incorrect `this.constructor` pattern
+  promise = new Promise !-> it 42
+  promise@@ = FakePromise1 = !-> it ->, ->
+  assert.ok promise.finally(->) instanceof Promise, 'subclassing, incorrect `this` pattern'
+  # NewPromiseCapability validations
+  promise = new Promise !-> it 42
+  promise@@ = FakePromise1 = !-> it ->, ->
+  FakePromise1[Symbol?species] = !->
+  assert.throws (!-> promise.finally(->)), 'NewPromiseCapability validations, #1'
+  FakePromise1[Symbol?species] = !-> it null, ->
+  assert.throws (!-> promise.finally(->)), 'NewPromiseCapability validations, #2'
+  FakePromise1[Symbol?species] = !-> it ->, null
+  assert.throws (!-> promise.finally(->)), 'NewPromiseCapability validations, #3'
+  #
+  # when value is returned, test finally is called, then onFulfilled is reached, not onRejected
+  # js:
+  # var finallyReached1 = false;
+  # var expectedValue1 = 1;
+  # var p1 = new Promise((resolve, reject) => resolve(expectedValue1));
+  # p1 = p1.finally(() => finallyReached1 = true;);
+  # p1.then(
+  #   val => {
+  #     assert.equal(finallyReached1, true);
+  #     assert.equal(value, expectedValue1);
+  #   },
+  #   reason => { assert.fail('onRejected should not be reached') });
+  #
+  # when exception is thrown, test finally is called, then onRejected is reached, not onFulfilled
+  # js:
+  # var finallyReached2 = false;
+  # var expectedReason1 = 1;
+  # var p2 = new Promise((resolve, reject) => reject(expectedReason1));
+  # p2 = p2.finally(() => finallyReached2 = true;);
+  # p2.then(
+  #   val => { assert.fail('onFulfilled should not be reached') },
+  #   reason => {
+  #     assert.equal(finallyReached2, true);
+  #     assert.equal(reason, expectedReason1);
+  #   });
+  #
+  # when onFinally is not a function nothing untoward happens
+  # var expectedValue2 = 1;
+  # var p1 = new Promise((resolve, reject) => resolve(expectedValue2));
+  # p1 = p1.finally(3);
+  # p1.then(
+  #   val => {
+  #     assert.equal(value, expectedValue2);
+  #   },
+  #   reason => { assert.fail('onRejected should not be reached') });
+  # var expectedReason2 = 1;
+  # var p2 = new Promise((resolve, reject) => reject(expectedReason2));
+  # p2 = p2.finally(3);
+  # p2.then(
+  #   val => { assert.fail('onFulfilled should not be reached') },
+  #   reason => {
+  #     assert.equal(reason, expectedReason2);
+  #   });
+  #
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Changelog
 ##### Unreleased
 - Added `Math.{clamp, deg-per-rad, degrees, fscale, rad-per-deg, radians, scale}`, [stage 1 proposal](https://github.com/rwaldron/proposal-math-extensions), [#226](https://github.com/zloirock/core-js/issues/226)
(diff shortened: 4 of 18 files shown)


library/modules/es7.promise.finally.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+// https://github.com/tc39/proposal-promise-finally
+var $export = require('./_export');
+
+$export($export.P, 'Promise', {
+  'finally': function _finally(onFinally){
```
clavecoder: Used `_finally`. Error generated without the underscore.

modules/es7.promise.finally.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+// https://github.com/tc39/proposal-promise-finally
+var $export = require('./_export');
+
+$export($export.P, 'Promise', {
+  'finally': function _finally(onFinally){
```
clavecoder: Used `_finally`. Error generated without the underscore.

tests/tests/es7.promise.finally.ls
```diff
@@ -0,0 +1,29 @@
+{module, test} = QUnit
+module \ES7
+
+test 'Promise#finally' (assert)!->
+  assert.isFunction Promise::finally
+  NATIVE and assert.arity Promise::finally, 2 # FF - 0
+  assert.name Promise::finally, \_finally
```
clavecoder: Remove underscore when name is fixed to `finally` from `_finally`.

modules/es7.promise.finally.js
Kovensky:
```js
{
  finally (onFinally) {
    ...
  }
}
```

modules/es7.promise.finally.js
```diff
@@ -0,0 +1,27 @@
...
+    var handler = typeof onFinally === 'function' ? onFinally : function(){};
+    var C = Promise;
+    var newPromise = Promise.prototype.then.call(
+      this, // throw if IsPromise(this) is not true
+      function(x) {
+        return new C()(function(resolve) {
```
Kovensky: You're calling the result of the `new C()` constructor as a function o_O

modules/es7.promise.finally.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+// https://github.com/tc39/proposal-promise-finally
+var $export = require('./_export');
+
+$export($export.P, 'Promise', {
+  'finally': function _finally(onFinally){
+    var handler = typeof onFinally === 'function' ? onFinally : function(){};
+    var C = Promise;
```
Kovensky: You're hardcoding the constructor as Promise instead of the species constructor

library/modules/es7.promise.finally.js
zloirock: Use `'finally': function(onFinally)`. For ES6+ engines it will create the correct `.name`, for old engines we can't do anything.

tests/tests/es7.promise.finally.ls
zloirock: Remove this test.

fn/promise/index.js
```diff
@@ -0,0 +1,2 @@
+require('../../modules/es7.promise.finally');
+module.exports = require('../../modules/_core').Promise;
```
zloirock: We already have `fn/promise.js`. Also, here missed `es6.promise`.

modules/es7.promise.finally.js
zloirock: @Kovensky we haven't method shorthands in ES3 / ES5.

modules/es7.promise.finally.js
zloirock: Use `_species-constructor`. Also, what is `Promise`? We can't use this global variable, if you need it for something - you can use `core.Promise || global.Promise`.

tests/tests/es7.promise.finally.ls
```diff
@@ -0,0 +1,29 @@
...
+  promise = new Promise !-> it 42
+  promise@@ = FakePromise1 = !-> it ->, ->
+  FakePromise1[Symbol?species] = !->
+  assert.throws (!-> promise.finally(->)), 'NewPromiseCapability validations, #1'
+  FakePromise1[Symbol?species] = !-> it null, ->
+  assert.throws (!-> promise.finally(->)), 'NewPromiseCapability validations, #2'
+  FakePromise1[Symbol?species] = !-> it ->, null
+  assert.throws (!-> promise.finally(->)), 'NewPromiseCapability validations, #3'
+
+
```
zloirock: Too many empty lines. Also, where `tests/library/es7.promise.finally.ls`?

tests/tests/es7.promise.finally.ls
```diff
@@ -0,0 +1,29 @@
+{module, test} = QUnit
+module \ES7
+
+test 'Promise#finally' (assert)!->
+  assert.isFunction Promise::finally
+  NATIVE and assert.arity Promise::finally, 2 # FF - 0
```
zloirock: `FF - 0`?

tests/tests/es7.promise.finally.ls
```diff
@@ -0,0 +1,29 @@
...
+  assert.ok promise.finally(->) instanceof Promise, 'subclassing, incorrect `this` pattern'
+  # NewPromiseCapability validations
+  promise = new Promise !-> it 42
+  promise@@ = FakePromise1 = !-> it ->, ->
+  FakePromise1[Symbol?species] = !->
+  assert.throws (!-> promise.finally(->)), 'NewPromiseCapability validations, #1'
+  FakePromise1[Symbol?species] = !-> it null, ->
+  assert.throws (!-> promise.finally(->)), 'NewPromiseCapability validations, #2'
+  FakePromise1[Symbol?species] = !-> it ->, null
+  assert.throws (!-> promise.finally(->)), 'NewPromiseCapability validations, #3'
```
zloirock: I see only subclassing tests. For `Promise#then` it's enough because of the main test case for them - https://github.com/promises-aplus/promises-tests, but `finally` missed here.

modules/es7.promise.finally.js
clavecoder: That's the behavior of the polyfill:
https://github.com/tc39/proposal-promise-finally/blob/84cec8005364a0cae413cad3be1e47b6a26be369/polyfill.js#L33

Kovensky: It crashes at runtime 🙀 

I've asked the author about it, but it's most likely accidental leftovers from refactoring.

I am using a previous version of the polyfill that had a much simpler `speciesConstructor` implementation (https://github.com/tc39/proposal-promise-finally/blob/c626fb5857e0564eff6ab6a5b5e2b9383c156ba7/polyfill.js).

zloirock: So maybe it's a bug of this polyfill? It can't work.

clavecoder: @kovensky:

1. Thanks for the older implementation reference. That's what I would expect. I like to move methodically through uncharted waters.
1. I'll use the core-js internal implementation of `speciesConstructor` used in `es6.promise.js`
1. Thanks for the quick responses!

Kovensky: It behaves correctly without the excess `()`.

modules/es7.promise.finally.js
clavecoder: That's why my second open item is asking the right way in core-js to get the species constructor. I think I've found it now (not checked in yet) but I'm still getting `TypeError` with `delay` (see image in initial comment).

tests/tests/es7.promise.finally.ls
clavecoder: Left over from cut/paste of `then`. Removed (not checked in yet).

tests/tests/es7.promise.finally.ls
clavecoder: Extra lines removed and `tests/library/es7.promise.finally.ls` added with tests copied from `then` (not checked in yet). Working on real tests...

fn/promise/index.js
clavecoder: > We already have `fn/promise.js`
OK, `require`d es7.promise.finally in `fn/promise.js` in leu of `fn/promise/finally.js`. Or do you want them both, so they can just add finally?

tests/tests/es7.promise.finally.ls
clavecoder: Done (not checked in  yet...).

library/modules/es7.promise.finally.js
clavecoder: > Use 'finally': function(onFinally)

Done (not checked in yet...)

clavecoder:
Checked in changes per comments. However my added test file is failing hard. Changes made:

1. `es7/promise.js`: Added require of `es6.promise` before `es7.promise.finally`
1. `fn/promise.js`: Added require of `es7.promise.finally` after `es6.promise`
1. Removed `fn/promise/` in favor of `fn/promise.js`
1. `library/es7/promise.js`: Added require of `es6.promise` before `es7.promise.finally`
1. `library/modules/es7.promise.finally.js`:
   1. changed to implicit name for `finally`
   1. now uses `speciesConstructor`
   1. Removed polyfill regression of using Promise instance as function.
1. `modules/es7.promise.finally.js`: *ditto*
1. Added `/tests/library/es7.promise.finally.ls`
   1. Mirrored behavior of `/tests/tests/es7.promise.finally.ls`
1. `/tests/tests/es7.promise.finally.ls`
   1. Removed extraneous comment
   1. Removed test for name of Promise#finally
   1. Added JavaScript pseudocode for Promise#finally implementation tests.

Open issues:
- [ ] `tests/library.js` is failing because `Promise#finally` is not found for test file `/tests/library/es7.promise.finally.ls`. I've tried tracing the `require` path and, of course, it seems right, but there's obviously something wrong. Also remember that the implementation was changed to use an anonymous function for the property assignment. I don't know if it is related. Error message:
image
The source code for the error line is `/tests/library/es7.promise.finally.ls/L7`
```
  assert.isFunction Promise::finally
```
- [ ] I'm finding LiveScript a bit beyond me for creating the unit tests without an example. I've added JavaScript pseudocode in comments  for `/tests/library/es7.promise.finally.ls` and `/tests/tests/es7.promise.finally.ls`. If you can point me to some LiveScript tests that are doing something similar, I should be able to tease out the implementation.

library/modules/es7.promise.finally.js
clavecoder: Checked in.

tests/tests/es7.promise.finally.ls
clavecoder: Checked in.

modules/es7.promise.finally.js
clavecoder: Checked in.

modules/es7.promise.finally.js
clavecoder: Checked in.

fn/promise/index.js
clavecoder: Checked in. Note that `fn/promise/` has been removed.

tests/tests/es7.promise.finally.ls
clavecoder: Checked in. Only have pseudocode. :frowning: Can you point me to a similar example?

tests/tests/es7.promise.finally.ls
clavecoder: Checked in.

tests/tests/es7.promise.finally.ls
clavecoder: Checked in. Only have pseudocode. :frowning: Can you point me to a similar example?

clavecoder:
Any test code you can point me to that work similarly to the pseudocode? I should have more time this weekend.

cvbuelow:
Is this still being worked on?

zloirock:
https://github.com/zloirock/core-js/commit/ed59a9c0822eb1fa3509e06f996f424d8392d0a9, https://github.com/zloirock/core-js/commit/86ed553fbaa2d1e2f380bfe174f158e711e2e51f, https://github.com/zloirock/core-js/commit/fd70e24f8f076177c321b29aa2dc3b4de8592122, https://github.com/zloirock/core-js/commit/0bc7a9ea75fd1a604cc6c99eb755c32a08afda9a

PR #252, opened 2016-10-18
Wrap frozen keys fix in try-catch not to break other WeakMap polyfills implementations

eugef:
Other WeakMap polyfills throw an error if frozen object is being passed as a key.
So we simply ignore the error that this check produces.

This could happen if another polyfill was already executed on the page.

### DIFF
--- a/modules/es6.weak-map.js
+++ b/modules/es6.weak-map.js
@@ -36,21 +36,23 @@ var methods = {
 var $WeakMap = module.exports = require('./_collection')('WeakMap', wrapper, methods, weak, true, true);
 
 // IE11 WeakMap frozen keys fix
-if(new $WeakMap().set((Object.freeze || Object)(tmp), 7).get(tmp) != 7){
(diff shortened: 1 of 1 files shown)


eugef:
@zloirock could you review the PR?

Kovensky:
If there is a WeakMap polyfill that will crash when receiving a frozen object, then the bad polyfill _should_ get overridden by core-js, instead of crashing (or being skipped, like this PR does).

Something like `function isBad () { try { return new $WeakMap().set((Object.freeze || Object)(tmp), 7).get(tmp) != 7 } catch (e) { return true } } if (isBad()) { ... }` might be more appropriate.

eugef:
@Kovensky you are right, but unfortunately it is not always possible to do that. 
And my PR just allows `core-js` to be more friendly to other code that can be already used on the page.

zloirock:
I agree with @Kovensky. I'll add something for that.

PR #253, opened 2016-10-28
Update es6.array.find-index.js

OleksiiZubko:
Looks like findIndex should return -1 (http://www.ecma-international.org/ecma-262/6.0/#sec-array.prototype.findindex)

### DIFF
--- a/library/modules/es6.array.find-index.js
+++ b/library/modules/es6.array.find-index.js
@@ -8,7 +8,8 @@ var $export = require('./_export')
 if(KEY in [])Array(1)[KEY](function(){ forced = false; });
 $export($export.P + $export.F * forced, 'Array', {
   findIndex: function findIndex(callbackfn/*, that = undefined */){
-    return $find(this, callbackfn, arguments.length > 1 ? arguments[1] : -1);
+    var result = $find(this, callbackfn, arguments.length > 1 ? arguments[1] : undefined);
+    return result || -1;
   }
 });
 require('./_add-to-unscopables')(KEY);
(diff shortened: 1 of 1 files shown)


Kovensky:
This will return `-1` if the result is `0`.

The `$find` helper itself should already return `-1` when the result is not found: https://github.com/OleksiiZubko/core-js/blob/df35f75b5f819951a8d900f3aed5e032217dcea1/library/modules/_array-methods.js#L42

OleksiiZubko:
@Kovensky Thanks, I've got. Then the issue is in documentations, because I'm using core-js v0.9.7 and documentations says:

`/**
      * Returns the value of the first element in the array where predicate is true, and undefined
      * otherwise.
      * @param predicate find calls predicate once for each element of the array, in ascending
      * order, until it finds one where predicate returns true. If such an element is found, find
      * immediately returns that element value. Otherwise, find returns undefined.
      * @param thisArg If provided, it will be used as the this value for each invocation of
      * predicate. If it is not provided, undefined is used instead.
      */
    find(predicate: (value: T, index: number, obj: Array<T>) => boolean, thisArg?: any): T;

    /**
      * Returns the index of the first element in the array where predicate is true, and undefined
      * otherwise.
      * @param predicate find calls predicate once for each element of the array, in ascending
      * order, until it finds one where predicate returns true. If such an element is found, find
      * immediately returns that element value. Otherwise, find returns undefined.
      * @param thisArg If provided, it will be used as the this value for each invocation of
      * predicate. If it is not provided, undefined is used instead.
      */
    findIndex(predicate: (value: T) => boolean, thisArg?: any): number;`

Kovensky:
Yeah, it looks like the docs were copypasted :D

OleksiiZubko:
yes :D . I'll close this PR, will create an issue for that. Thank you

Issue #119, opened 2015-10-08
Error with Getters and Setters: Object Expected (IE8)


bmakuh:
Right now I `require('core-js')` via Browserify-rails, but then when I load up the page in IE8, I get an error (Object Expected) on this line:

```js
if('get' in Attributes || 'set' in Attributes)throw TypeError('Accessors not supported!');
```

zloirock:
You uses `Object.defineProperty` / `Object.defineProperties` / `Object.create` with getters / setters. In ES3 is no way to add accessors, IE8 supports it only on DOM objects, so we generate this error.

Duplicate #84, #74.

txiaocao:
i has some problem.how to do?

blling:
What is the solution?  one of a 3rd lib use it :(

zloirock:
@blling there is no solution. You can propose this change to the 3rd library, replace this library to something else or stop support IE8.

Issue #84, opened 2015-06-22
Error thrown in IE8


callumacrae:
I'm using core-js as a dependency of babel, and I'm getting the following error thrown in IE8:

> Exception thrown and not caught

Thanks, IE8.

It's coming from this line, so I'm assuming that this is the exception it's talking about: https://github.com/zloirock/core-js/blob/master/modules/es5.js#L43

Any ideas? Doesn't happen in modern browsers, just IE8.

callumacrae:
dupe https://github.com/zloirock/core-js/issues/74#issuecomment-107828071

Issue #74, opened 2015-06-02
IE8 and FF3.6.28 parse error for core-js/es5


seeden:
when I use core-js/es5 I can see error in the console (IE8, ff 3.6.28): Accessors not supported.

The problem is with this line:
if('get' in Attributes || 'set' in Attributes)throw TypeError('Accessors not supported!');

Do you have any advice how to avoid it?
Thanks a lot

zloirock:
Looks like you uses `Object.defineProperty` / `Object.defineProperties` / `Object.create` with getters / setters. In ES3 is no way to add assessors, so we generate this error.

PR #258, opened 2016-11-04
Update _object-dp.js

skcary:
bugfixed on ie8


zloirock:
Sorry, a error should be thrown here in IE8. IE8 supports accessors on DOM elements (see the special case some lines above), but not on usual objects.

zloirock:
BTW duplicate of #74, #84, #119.

Issue #262, opened 2016-12-08
Importing Symbol properties before Symbol from core-js/library breaks Symbol


taion:
In a fresh session, run:

```js
require('core-js/library/fn/symbol/async-iterator');
typeof require('core-js/library/fn/symbol');
```

The latter gives `'object'`.

By contrast, if you just do

```js
typeof require('core-js/library/fn/symbol');
```

This gives `function` as expected.

This causes https://github.com/babel/babel/issues/4783.

PR #267, opened 2016-12-21
Don't set Symbol when loading asyncIterator

taion:
Fixes https://github.com/zloirock/core-js/issues/262, though I'm not sure how to test that.

Does this look right? This code doesn't seem to need to change `core.Symbol` if `core.Symbol` doesn't already exist, and the module building `core.Symbol` will pick up everything defined on `wks.store` anyway.

### DIFF
--- a/library/modules/_wks-define.js
+++ b/library/modules/_wks-define.js
@@ -4,6 +4,9 @@ var global         = require('./_global')
   , wksExt         = require('./_wks-ext')
   , defineProperty = require('./_object-dp').f;
 module.exports = function(name){
-  var $Symbol = core.Symbol || (core.Symbol = LIBRARY ? {} : global.Symbol || {});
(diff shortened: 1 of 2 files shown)


taion:
Does this look like it's on the right track?

taion:
Any updates here?

taion:
BTW, if we merge this, we should possibly revert https://github.com/babel/babel/pull/5195, since that workaround would no longer be relevant.

zloirock:
@taion this PR is not a completely correct fix - it will work only in some cases. Most likely, need to rewrite `_export`. I have a problem with time, but I'll try to do it near days.

taion:
My apologies then. There's no urgency here on my part any more, BTW – https://github.com/babel/babel/pull/5195 works around this adequately and even leads to slightly terser generated code on the Babel side.

taion:
I'm going to close this given that the comment above that the implementation isn't correct. The workaround on Babel is sufficient for my purposes, plus we still have the tracking issue anyway.

Issue #280, opened 2017-02-08
padStart/padEnd in current safari is buggy and shouldnt be used


trashhalo:
The currently released version of safari has a bug in padStart padEnd that causes it to return null. 
> Fixed a bug where String.prototype.padStart() and padEnd() would insert null in rare cases

https://developer.apple.com/safari/technology-preview/release-notes/ fix in preview 15

While it says its rare we are hitting it about 40% of the time. As far as I can tell there isn't a externally facing way to force a polyfill to override native. ~~Can you guys mark padStart/end as $export.F until this fix lands?~~

benjie:
I was hitting this also on iOS (weirdly it doesn't happen so much when the Developer Tools are connected, but much more when they are not); lost over an hour to finding out what the issue in my code was (not least because I couldn't debug it with the devtools and it was Mobile Safari so I had to resort to `alert()` based debugging like some kind of Neanderthal).

A buggyfill for this would have saved me a lot of time.

zloirock:
Is forced replacement for `/Version\/10\.\d+(\.\d+)? Safari\//.test(userAgent)` enough, I hope?

zloirock:
Fixed in `core-js@2.5.3`.

AlgusDark:
@zloirock 
I believe we need to change the regex and add `mobile`:
`/Version\/10\.\d+(\.\d+)? (Safari|Mobile)\//.test(userAgent)`

I'm facing the `buggy` problem in `Mobile` iphone.

sunderls:
safari: `Mozilla/5.0 (iPhone; CPU iPhone OS 10_2 like Mac OS X) AppleWebKit/602.3.12 (KHTML, like Gecko) Version/10.0 Mobile/14C89 Safari/602.1`

webview: `Mozilla/5.0 (iPhone; CPU iPhone OS 10_2 like Mac OS X) AppleWebKit/602.3.12 (KHTML, like Gecko) Mobile/14C89 Safari Line/7.16.0`

maybe we better check `/iphone os 10/i`?

zloirock:
@sunderls it will not work with desktop `Safari`. Feel free to propose better detection.

PR #281, opened 2017-02-08
Safari's impl of padStart/end is buggy. Force use of poly till fix released

trashhalo:
There is a bug in current safari that causes padStart/End to return null. https://bugs.webkit.org/show_bug.cgi?id=161944 its resolved in preview 15. I propose marking the core-js poly as forced until implementation stabilizes. Also the build changed a bunch of files. I am not sure if I am supposed to include those. First commit is what I changed second is what the build system changed.

PR for #280


zloirock:
Forced replacement of built-ins without testing is not a solution. If you can add a simple detection of broken feature, it should solve the problem.

trashhalo:
This bug isnt really detectable unless we are going to get into browser version sniffing. Im going to go ahead and close this pr

PR #284, opened 2017-02-18
Update .npmignore

aesopwolf:


zloirock:
Hmm... Ok, let's save some KB.

PR #286, opened 2017-03-09
Add support for Sphere Dispatch API

fatcerberus:
[miniSphere](https://github.com/fatcerberus/minisphere) is JavaScript game engine maintained by myself and using the Duktape ES5 engine.  My environment doesn't have any of the async harnesses (setTimeout, postMessage, etc.) supported by browsers, but *does* have a function, `Dispatch.now()`, that serves the same purpose.  the core-js Promise implementation doesn't currently work in my environment because of this; this pull adds support for it.

### DIFF
--- a/modules/_task.js
+++ b/modules/_task.js
@@ -7,6 +7,7 @@ var ctx                = require('./_ctx')
   , setTask            = global.setImmediate
   , clearTask          = global.clearImmediate
   , MessageChannel     = global.MessageChannel
+  , Dispatch           = global.Dispatch
   , counter            = 0
   , queue              = {}
   , ONREADYSTATECHANGE = 'onreadystatechange'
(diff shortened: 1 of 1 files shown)


zloirock:
Hmm... Ok, makes sense.

modules/_task.js
```diff
@@ -41,6 +42,11 @@ if(!setTask || !clearTask){
     defer = function(id){
       process.nextTick(ctx(run, id, 1));
     };
+  // Sphere (JS game engine) Dispatch API
+  } else if(Dispatch){
```
zloirock: Please, use `Dispatch && Dispatch.now` for preventing possible conflicts.

fatcerberus:
Okay, made the requested change.  Anything I need to do?

zloirock:
Nope, thanks.

PR #290, opened 2017-03-21
Remove non compliant code

pierrepinon:
In this case function declaration is better than Function constructor

Source :
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function
- http://stackoverflow.com/questions/3799295/function-constructor-vs-function-statement


loganfsmyth:
What is noncompliant about this?

1. Your code is wrong because you've converted it from an IIFE to a function that is never invoked.
2. If it was an IIFE, this would change the behavior because you've moved the code from non-strict mode to strict mode.

PR #309, opened 2017-06-13
Fix array buffer with missing or fractional length

nicolo-ribaudo:
This is the behavior described in the specification

(I couldn't get the tests run on my PC, I hope I got them right)


nicolo-ribaudo:
This bug doesn't allow using node 8.x's "v8" module when the global "babel-polyfill" is loaded

zloirock:
Seems fixed in https://github.com/zloirock/core-js/commit/13920872c09438d97e5f20560740d2c9a9dfd230

PR #311, opened 2017-06-25
Put "++id" in parens

sebm:
Some minifiers (well, just JSMin) produce broken output if they attempt to minify an expression in the form `a + ++b`. This pull request addresses this by wrapping core-js's one such expression in parentheses.

Unfortunately there is currently no eslint rule to cover this except for "no-plusplus", which would require more extensive changes to this code.


loganfsmyth:
Isn't this just an argument to fix that minifier?

sebm:
I'd agree with you, but JSMin considers this a feature - see http://www.crockford.com/javascript/jsmin.html

loganfsmyth:
Easiest to switch to Uglify then? JS is a well-defined language, if a minifier can't preserve the semantics of the language that is a failure on their part, whether they see it that way or not, expecting third-party code to handle bugs in unrelated tools seems like a mistake.

I get that `core-js` is one specific case of this and you are right that this is a potential fix, but I don't see how you can trust a tool that decides to quietly break things like this. It should at least fail loudly when  it encounters code like this instead of formatting it incorrectly.

Consider that literally anything anywhere in your `node_modules` could have this exact same issue, or could introduce code like this at any time. I'd personally be terrified to have that risk.

zloirock:
Sorry, but I agree with @loganfsmyth.

PR #357, opened 2017-11-16
Update _redefine.js

collierrgbsitisfise:
MB smtg like this.....


zloirock:
`try / catch` block here is not required. Please, create an issue with maximum information and a reproducible example.

PR #364, opened 2017-12-07
Fix error related with wrong function call

mdasmendel:
Fix error Function.prototype.toString: 'this' is not a Function object

Sorry I forgot the main error but It work well now for me (from my fork), so I awaiting some fixes for switch to original repository


zloirock:
Please, add info about the error. Anyway, it should be fixed another way than that.

zloirock:
It's related #356 / #357, so adding an issue with a reproducible example could be useful.

PR #367, opened 2017-12-14
[Feedback needed] fix; exit all manually entered NodeJS domains

aheckmann:
I've found an issue with core-js Promises when used in Node 4x. Using the following script demonstrates that when [domains](https://nodejs.org/docs/latest-v4.x/api/domain.html) are in use and a `then()` callback throws an error, the active domain is left in an entered state. This causes the domain to not be garbage collected, leading to memory leaks of not just the domain, but of anything added to the domain (a pretty big leak when using them to handle http server errors as outlined in the domain documentation).

```js
'use strict';

var core = require('core-js/library');
var domain = require('domain');

var Type = {
  'core': core.Promise,
  'native': Promise
};

var type = process.argv[2]; // `node thisfile.js core` or `node thisfile.js native`
console.log('using %s', type);
var P = Type[type];

function reportDepth(state) {
  console.log(state + ', current domain depth is ' + domain._stack.length);
}

var resolver;
var d = domain.create();
// d.add(somethingLarge);

d.run(function() {
  reportDepth('In domain.run()');

  new P(function(resolve) {
    resolver = resolve;
  }).then(function() {
    reportDepth('Prior to throwing an error');

    throw new Error('Cause the domain to be left in an entered state');
  }).catch(function() {
    reportDepth('In exception handler');
  });
});

reportDepth('After domain exit');
resolver('asynchronous resolve()');

process.on('exit', function() {
  reportDepth('At exit');
});
```

This PR fixes the issue I'm facing but I'm not sure how / where to write tests for this project. I'm also not sure if I need to update both `modules/es6.promise.js` and `library/modules/es6.promise.js`. Please advise.

### DIFF
--- a/library/modules/es6.promise.js
+++ b/library/modules/es6.promise.js
@@ -53,7 +53,7 @@ var notify = function (promise, isReject) {
       var resolve = reaction.resolve;
       var reject = reaction.reject;
       var domain = reaction.domain;
-      var result, then;
+      var result, then, exited;
       try {
         if (handler) {
           if (!ok) {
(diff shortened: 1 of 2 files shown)


zloirock:
LGTM

PR #382, opened 2018-02-20
[CI] Update node versions

nicolasleger:


zloirock:
I'm not sure that it's really required, but OK, I'll accept it.

PR #397, opened 2018-05-03
fix #396

### LINKED ISSUE #396: ES6 Promise polyfill failing in some environments
Hello,

So, I'm developing an app for LG WebOS 2. I write it in ES6, build it with webpack and transpile it to ES5 with babel loader. I'm using babel polyfill like this

```
import "babel-polyfill";
```

in the beginning of the app. Everything is standard so far.

I'm aware of the fact that most people probably don't develop for WebOS so let me just say that it's a regular modern web browser and supports everything a regular browser supports.

In normal app there is no problem. But I'm trying to use web workers, which WebOS supports. My web worker code uses some modules I've written for the main app and those modules use stuff like promises, generators, etc. I want to use those the same way I'd use them in the main app so that's why I'm also using babel-polyfill in the web worker code. And here is the problem.

When I run the web worker, it throws an error

> TypeError: Expected at least one argument

on [this line](https://github.com/zloirock/core-js/blob/master/modules/_microtask.js#L43).

```
// modules/_microtask.js line 43
var promise = Promise.resolve();
```

I'm not exactly sure why but there seems to a problem with the fact that when Promise is polyfilled, Promise.resolve is defined as a function that expects one argument [here](https://github.com/zloirock/core-js/blob/master/modules/es6.promise.js#L203).

```
// modules/es6.promise.js line 203
OwnPromiseCapability = function () {
    var promise = new Internal();
    this.promise = promise;
    this.resolve = ctx($resolve, promise, 1);
    this.reject = ctx($reject, promise, 1);
  };
```
while on the line that it failed it's called without any arguments. It's probably a property of the environment in which the web worker runs on WebOS. When I explicitly changed the line in my built code to
```
var promise = Promise.resolve(undefined);
````
it works.

Maybe the line should be changed to this to prevent errors like this in exotic environments. I don't see any other solution.

kytart:
This pull request fixes the problem I described in issue #396.

### DIFF
--- a/modules/_microtask.js
+++ b/modules/_microtask.js
@@ -40,7 +40,7 @@ module.exports = function () {
     };
   // environments with maybe non-completely correct, but existent Promise
   } else if (Promise && Promise.resolve) {
-    var promise = Promise.resolve();
+    var promise = Promise.resolve(undefined);
     notify = function () {
       promise.then(flush);
     };


zloirock:
Ok, thanks.

PR #400, opened 2018-05-10
[Request] Rename variable name

ilyanew:
Rename variable 'final' to 'fin'. Because 'final' is reserved word in javascript. And it breaks down 'yui compressor'.
Please find the list of reserved words in js:
https://www.w3schools.com/js/js_reserved.asp
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar


loganfsmyth:
Not the maintainer of this project, but I'll just say that `final` hasn't been a reserved word since 2009 when the ES5 spec was released, so at a minimum this is still a bug with `yui compressor` for any modern code.

ilyanew:
Yeap, You are absolutely right. 'Yui compressor' is too old :) 
But if you merge this PR, 'core-js' can be easily transformed to es3. I believe that it will only a benefit for 'core-js'. 

I would be really grateful!
You would have helped me out very much.

zloirock:
Please, leave only [this](https://github.com/zloirock/core-js/pull/400/commits/7f809e5f94dd05e5e37b77644049a720080c1d92) commit and I'll merge it.

ilyanew:
There are some issues with newlines in a few files. :)
But I changed variable name only in 'modules/es6.typed.array-buffer.js' file.
Or should I create new PR?

zloirock:
Just do `git rebase -i` and drop the rest commits.

ilyanew:
Sorry, I've faced some problems with Rebasing. Therefore decided to create new PR https://github.com/zloirock/core-js/pull/402

zloirock:
Ok, thanks -)

zloirock:
I you wanna contribute to this project, you can try to improve linting settings for prevent problems like this in the future.

PR #402, opened 2018-05-14
Update es6.typed.array-buffer.js

ilyanew:
Rename variable 'final' to 'fin'. Because 'final' is reserved word in javascript. And it breaks down 'yui compressor'.
Please find the list of reserved words in js:
https://www.w3schools.com/js/js_reserved.asp
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar


ilyanew:
This PR was created instead of https://github.com/zloirock/core-js/pull/400

PR #409, opened 2018-06-08
Mention polyfill.bundle.run in README

xtuc:
Since it's using your polyfills I thought It would be great to mention it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,8 @@ require('core-js');
 ```
 If you need already bundled version of `core-js`, use `core-js-bundle` `npm` package or a [version of this package from CDN](https://unpkg.com/core-js-bundle@3.0.0-beta.3) ([minified version](https://unpkg.com/core-js-bundle@3.0.0-beta.3/minified.js)).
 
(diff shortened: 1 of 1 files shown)


zloirock:
I'm not sure that actual approach makes sense. Usage CDN with generation required polyfill definitely is a good idea, but not bundling of each polyfill to a separate bundle. In this case, we will have duplication of internal modules in each bundle. If dozens of polyfills are required, together they will have a monstrous size.

xtuc:
Yes, but having seperated plugins allows people to pollyfill only one API. Iterating on DOM nodes for example is quite common. 

I'm planning to add support for babel-preset-env in bundle.run, which would allow people to select (and bundle) every plugins for a given target at once, would it make sense to you?

zloirock:
> Yes, but having seperated plugins allows people to pollyfill only one API. Iterating on DOM nodes for example is quite common.

I'm not sure that is enough common case.

> I'm planning to add support for babel-preset-env in bundle.run, which would allow people to select (and bundle) every plugins for a given target at once, would it make sense to you?

Yeh, what's really good conception.

PR #411, opened 2018-06-13
Make RegExp#[Symbol.*] methods call exec

nicolo-ribaudo:
Ref: https://github.com/babel/babel/pull/7105#discussion_r194269147


zloirock:
Sorry, some more days I'll be unavailable. I'll review it later.

packages/core-js/internals/fix-regexp-well-known-symbol-logic.js
```diff
@@ -16,13 +16,17 @@ module.exports = function (KEY, length, exec) {
...
       // 21.2.5.8 RegExp.prototype[@@replace](string, replaceValue)
       // 21.2.5.11 RegExp.prototype[@@split](string, limit)
       ? function (string, arg) { return regexMethod.call(string, this, arg); }
       // 21.2.5.6 RegExp.prototype[@@match](string)
       // 21.2.5.9 RegExp.prototype[@@search](string)
       : function (string) { return regexMethod.call(string, this); }
     );
+    // TODO: This line makes the tests fail:
+    //       ReferenceError: Can't find variable: Reflect
+    // hide(RegExp.prototype[SYMBOL], 'name', '[Symbol.' + KEY + ']');
```
zloirock: In many engines, functions `.name` property non-writable non-configurable, so, I think, it would be better just remove it.

packages/core-js/modules/es.string.match.js
```diff
@@ -1,11 +1,42 @@
...
+    // `String.prototype.match` method
+    // https://tc39.github.io/ecma262/#sec-string.prototype.match
+    function match(regexp) {
+      var O = defined(this);
+      var matcher = regexp == undefined ? undefined : regexp[MATCH];
+      return matcher !== undefined ? matcher.call(regexp, O) : new RegExp(regexp)[MATCH](String(O));
+    },
+    // `RegExp.prototype[@@match]` method
+    // https://tc39.github.io/ecma262/#sec-regexp.prototype-@@match
+    function Symbol$match(regexp) {
```
zloirock: For better compression, use named functions only when they should have the same explicit `.name` property (like `match` above).

tests/helpers/helpers.js
```diff
@@ -54,3 +54,21 @@ export function timeLimitedPromise(time, fn) {
...
+export function patchRegExp$exec(run) {
+  return assert => {
+    const originalExec = RegExp.prototype.exec;
+    // eslint-disable-next-line no-extend-native
+    RegExp.prototype.exec = function (...args) {
+      return originalExec.apply(this, args);
+    };
+    try {
+      return run(assert);
+    } finally {
```
zloirock: IIRC in very old IE `try / finally` does not work without `catch`.

packages/core-js/internals/advance-string-index.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
```
zloirock: Please, add a comment that it's `AdvanceStringIndex` abstract operation and the link to the spec.

packages/core-js/internals/fix-regexp-well-known-symbol-logic.js
```diff
@@ -16,13 +16,17 @@ module.exports = function (KEY, length, exec) {
     return ''[KEY](O) != 7;
   })) {
```
zloirock: Are you sure that this test also covers all engines without internal execution of `.exec` method? Maybe makes sense extend it?

packages/core-js/modules/es.string.match.js
```diff
@@ -1,11 +1,42 @@
...
+      var S = String(this);
+
+      if (!rx.global) return rx.exec(S);
+
+      var fullUnicode = rx.unicode;
+      rx.lastIndex = 0;
+      var A = [];
+      var n = 0;
+      var result;
+      while ((result = rx.exec(S)) !== null) {
```
zloirock: In [`RegExpExec`](https://tc39.github.io/ecma262/#sec-regexpexec), own `.exec` should be called only if it's callable. Mabe makes sense to implement `RegExpExec` abstract operation?

packages/core-js/internals/fix-regexp-well-known-symbol-logic.js
```diff
@@ -5,24 +5,30 @@ var fails = require('../internals/fails');
...
+module.exports = function (KEY, length, exec, sham) {
   var SYMBOL = wellKnownSymbol(KEY);
   var methods = exec(requireObjectCoercible, SYMBOL, ''[KEY]);
   var stringMethod = methods[0];
   var regexMethod = methods[1];
   if (fails(function () {
+    var execCalled = false;
+    var re = /a/;
+    re.exec = function () { execCalled = true; return null; };
+    return re[SYMBOL] && (re[SYMBOL](''), execCalled);
```
zloirock: It should return a negative result for the existent feature. Use, for example,
```js
re[SYMBOL]('');
return !execCalled;
```

zloirock: It's just an example. The main idea of the comment - the result should be inverted.

packages/core-js/internals/fix-regexp-well-known-symbol-logic.js
```diff
@@ -5,24 +5,33 @@ var fails = require('../internals/fails');
...
   var methods = exec(requireObjectCoercible, SYMBOL, ''[KEY]);
   var stringMethod = methods[0];
   var regexMethod = methods[1];
   if (fails(function () {
+    // String methods call symbol-named RegEp methods
     var O = {};
     O[SYMBOL] = function () { return 7; };
     return ''[KEY](O) != 7;
+  }) || fails(function () {
+    // Symbol-named RegExp methods call .exec
```
nicolo-ribaudo: This test throws an uncaught stack overflow error in the promises tests and I'm not sure why.

nicolo-ribaudo:
@zloirock I manages to make the tests pass. Could you review this PR again?

DmitrySoshnikov:
Hey guys, is this PR good to go yet? Is there still anything blocking?

nicolo-ribaudo:
@zloirock Since Babel uses core-js@2, can I backport this PR?

zloirock:
@nicolo-ribaudo feel free to add the same PR to `v2` branch and I'll publish a new `core-js@2` version. But let's land this PR at first.

zloirock:
At first sight, LGTM.

zloirock:
@nicolo-ribaudo `RegExp#@@replace correctly handles substitutions` tests fails in many browsers - for example, in modern FF. Could you fix it?

nicolo-ribaudo:
How do I run the tests in a browser?

zloirock:
’tests/tests.html’25.09.2018, 12:18, "Nicolò Ribaudo" <notifications@github.com>:How do I run the tests in a browser?

zloirock:
@nicolo-ribaudo
Chrome 51:
image
Chrome 50:
image
image

nicolo-ribaudo:
I will fix it later.
Just for curiosity, do you have multiple versions of chrome installed on the same machine?

zloirock:
I use Saucelabs and run tests in all available browsers before releases.

zloirock:
IE8:
image

nicolo-ribaudo:
I'm working on this now

nicolo-ribaudo:
> @nicolo-ribaudo `RegExp#@@replace correctly handles substitutions` tests fails in many browsers - for example, in modern FF. Could you fix it?

Which firefox version? I can't reproduce it.

---

I'm writing down some notes for myself

- The tests are failing on IE8 because this code:

  ```js
  var re = /a/;
  re.exec("a");
  re.lastIndex;
  ```

  Should evaluate to 0 (it works in Chrome) but evaluates to 1 in IE8.

- The tests are failing on Chrome 51 because native split doesn't work with patched
  `RegExp#exec`: these two snippets should produce the same output but they don't.

  ```js
  "ab".split(/a*?/);

  > ["a", "b"]
  ```
  ```js
  var exec = RegExp.prototype.exec;
  RegExp.prototype.exec = function () { return exec.apply(this, arguments); };
  "ab".split(/a*?/);

  > ["ab"]
  ```

- The tests are failing on Chrome 50 because the polyfilled `RegExp#@@match`,
  since `.exec` hasn't been modified, calls `String#match` assuming that it
  won't call `RegExp#@@match` again (this assumption is wrong).

- Safari - TODO

Conclusion: I also need to polyfill `.exec` :man_facepalming:

zloirock:
@nicolo-ribaudo 
> Which firefox version? I can't reproduce it.

IIRC it was fixed in https://github.com/zloirock/core-js/pull/434

nicolo-ribaudo:
@zloirock In order to fix some IE bugs, I need to fix this IE bug:

```js
> /^(a?)/.exec("")
["", ""] // correct

> /^(a)?/.exec("")
> ["", ""] // wrong, should be ["", undefined]
```
I don't think that it is possible, unless we re-implement the regexp engine. Could I disable an assertion on IE8?

zloirock:
Available in `core-js@3.0.0-beta.4`.

PR #448, opened 2018-10-31
Fix syntax error in code example

dsonet:


zloirock:
Thanks for the PR, but, as @nicolo-ribaudo explained, it's not an error.

PR #465, opened 2018-12-17
Fix regression #463

### LINKED ISSUE #463: Out of stack space error in IE 11 when loading shim.min.js from core-js 2.6.0
I have an Angular app, and it works fine in all browsers when loading shim.min.js from core-js v2.5.7. However, after updating to core-js v2.6.0, I get an "Out of stack space" error in shim.min.js in IE 11.

I tried switching to the non-minified file and got the same error, on the following line:
https://github.com/zloirock/core-js/blob/f980318866f5cfb2933493c81ba89e1baf0f33c2/client/shim.js#L6312

nicolo-ribaudo:
Fixes #463


client/shim.js
```diff
@@ -6308,9 +6412,11 @@ __webpack_require__(59)('split', 2, function (defined, SPLIT, $split) {
     };
   // Chakra, V8
   } else if ('0'[$SPLIT](undefined, 0)[LENGTH]) {
-    $split = function (separator, limit) {
+    internalSplit = function (separator, limit) {
```
nicolo-ribaudo: This change in the compiled output is what actually fixes the bug.

PR #472, opened 2019-01-07
Fix #471

### LINKED ISSUE #471: Updates to String.prototype.replace doesn't handle $ in replacement value without capture group
The recent updates to String.prototype.replace are causing issues when a replacement without a capture group starts with a $. 

I was specifically seeing this in Safari 12 but didn't dig in on that front beyond knowing that it wasn't affecting Chrome.

```javascript
Chrome: '{price} Retail'.replace(/{price}/g, '$25.00') == '$25.00 Retail'
Safari: '{price} Retail'.replace(/{price}/g, '$25.00') == '5.00 Retail'
```

Version: 2.6.1

nicolo-ribaudo:
Fix #471

I will backport this PR tomorrow

### DIFF
--- a/packages/core-js/modules/es.string.replace.js
+++ b/packages/core-js/modules/es.string.replace.js
@@ -112,12 +112,12 @@ require('../internals/fix-regexp-well-known-symbol-logic')(
             break;
           default: // \d\d?
             var n = +ch;
-            if (n === 0) return ch;
+            if (n === 0) return match;
             if (n > m) {
               var f = floor(n / 10);
-              if (f === 0) return ch;
+              if (f === 0) return match;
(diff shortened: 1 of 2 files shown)


packages/core-js/modules/es.string.replace.js
```diff
@@ -112,12 +112,12 @@ require('../internals/fix-regexp-well-known-symbol-logic')(
             break;
           default: // \d\d?
             var n = +ch;
-            if (n === 0) return ch;
+            if (n === 0) return match;
```
nicolo-ribaudo: When the replacement is, for example `foo $12 bar`, `ch` is `12` while `match` is `$12`.

zloirock:
@nicolo-ribaudo IE:
image

Issue #372, opened 2018-01-24
Invalid flags supplied to RegExp constructor 'y' 


fernandogmar:
Hi Denis,

I expected that the RegExp polyfill would work with the sticky flag 'y'
`var identifierRe = new RegExp('[a-zA-Z_][a-zA-Z0-9_-]*', 'y');`

but in safari v9.1 it fails with error:
`Invalid flags supplied to RegExp constructor`

Am I doing something wrong? or doesn't this polyfill support the sticky flag at all?

Thanks in advance!

zloirock:
Hi. The sticky flag support is not polyfilled in `core-js`. After [this discussion](https://github.com/babel/babel/issues/904) I spent some time playing with the sticky flag but I didn't add it IIRC because of problems with some early implementations. At this moment, adding support for the sticky flag is not in my plans. Feel free to add a PR.

PR #492, opened 2019-03-07
Add support for the 'y' flag to the RegExp constructor

nicolo-ribaudo:
We already partially had this logic in `RegExp#@@split`: I removed it from there and moved to the `RegExp` exec internal polyfill.

The changes to the exec constructor are mainly moving things to a single expression to different statements, so that I could place the two `SUPPORTS_Y` blocks in between.

Fixes https://github.com/zloirock/core-js/issues/372

### DIFF
--- a/packages/core-js-compat/src/data.js
+++ b/packages/core-js-compat/src/data.js
@@ -791,7 +791,6 @@ module.exports = {
     chrome: '49',
     edge: '13',
     firefox: '3',
-    opera: '36',
     safari: '10.0',
   },
   'es.regexp.to-string': {
--- a/packages/core-js-compat/src/data.js
+++ b/packages/core-js-compat/src/data.js
@@ -787,6 +787,13 @@ module.exports = {
     firefox: '37',
     safari: '9.0',
   },
+  'es.regexp.sticky': {
+    chrome: '49',
+    edge: '13',
+    firefox: '3',
+    opera: '36',
+    safari: '10.0',
+  },
   'es.regexp.to-string': {
     chrome: '50',
     firefox: '46',
--- a/packages/core-js/internals/fix-regexp-well-known-symbol-logic.js
+++ b/packages/core-js/internals/fix-regexp-well-known-symbol-logic.js
@@ -1,5 +1,4 @@
 'use strict';
-var hide = require('../internals/hide');
 var redefine = require('../internals/redefine');
 var fails = require('../internals/fails');
 var wellKnownSymbol = require('../internals/well-known-symbol');
@@ -30,7 +29,7 @@ var SPLIT_WORKS_WITH_OVERWRITTEN_EXEC = !fails(function () {
   return result.length !== 2 || result[0] !== 'a' || result[1] !== 'b';
 });
 
-module.exports = function (KEY, length, exec, sham) {
+module.exports = function (KEY, length, exec) {
   var SYMBOL = wellKnownSymbol(KEY);
 
   var DELEGATES_TO_SYMBOL = !fails(function () {
@@ -88,6 +87,5 @@ module.exports = function (KEY, length, exec, sham) {
       // 21.2.5.9 RegExp.prototype[@@search](string)
       : function (string) { return regexMethod.call(string, this); }
     );
-    if (sham) hide(RegExp.prototype[SYMBOL], 'sham', true);
   }
 };
--- a/packages/core-js/internals/fix-regexp-well-known-symbol-logic.js
+++ b/packages/core-js/internals/fix-regexp-well-known-symbol-logic.js
@@ -19,6 +19,12 @@ var REPLACE_SUPPORTS_NAMED_GROUPS = !fails(function () {
   return ''.replace(re, '$<a>') !== '7';
 });
 
+// IE <= 11 replaces $0 with the whole match, as if it was $&
+// https://stackoverflow.com/questions/6024666/getting-ie-to-replace-a-regex-with-the-literal-string-0
+var REPLACE_KEEPS_$0 = (function () {
+  return 'a'.replace(/./, '$0') === '$0';
+})();
+
 // Chrome 51 has a buggy "split" implementation when RegExp#exec !== nativeExec
 // Weex JS has frozen built-in prototypes, so use try / catch wrapper
 var SPLIT_WORKS_WITH_OVERWRITTEN_EXEC = !fails(function () {
@@ -59,7 +65,7 @@ module.exports = function (KEY, length, exec) {
   if (
     !DELEGATES_TO_SYMBOL ||
     !DELEGATES_TO_EXEC ||
-    (KEY === 'replace' && !REPLACE_SUPPORTS_NAMED_GROUPS) ||
+    (KEY === 'replace' && !(REPLACE_SUPPORTS_NAMED_GROUPS && REPLACE_KEEPS_$0)) ||
     (KEY === 'split' && !SPLIT_WORKS_WITH_OVERWRITTEN_EXEC)
   ) {
     var nativeRegExpMethod = /./[SYMBOL];
@@ -74,7 +80,7 @@ module.exports = function (KEY, length, exec) {
         return { done: true, value: nativeMethod.call(str, regexp, arg2) };
       }
       return { done: false };
-    });
+    }, { REPLACE_KEEPS_$0: REPLACE_KEEPS_$0 });
     var stringMethod = methods[0];
     var regexMethod = methods[1];
 
--- a/packages/core-js/internals/regexp-exec.js
+++ b/packages/core-js/internals/regexp-exec.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var regexpFlags = require('./regexp-flags');
+var fails = require('./fails');
 
 var nativeExec = RegExp.prototype.exec;
 // This always refers to the native implementation, because the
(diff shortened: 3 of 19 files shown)


zloirock:
IIRC I had some problems with old FF when I played with something like this. Need to test it.

nicolo-ribaudo:
What do you mean approximately with "old"? I can try to download it and check.

zloirock:
I don't remember. Early implementations. It was in 2015.

zloirock:
Most likely it was related [this issue](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/sticky#Anchored_sticky_flag).

THEMAXIMUM77:
You suck

nicolo-ribaudo:
I have tested this PR in firefox 11: it was affected by that bug which I fixed in 775b17f.

`re.sticky` can't be set to true, because it is non-configurable and non-enumerble:
Schermata da 2019-03-13 00-30-50

packages/core-js/modules/es.regexp.constructor.js
```diff
@@ -30,13 +34,35 @@ if (FORCED) {
...
+      if (sticky) flags = flags.replace(/y/g, '');
+    }
+
+    var result = inheritIfRequired(
+      CORRECT_NEW ? new NativeRegExp(pattern, flags) : NativeRegExp(pattern, flags),
+      thisIsRegExp ? this : RegExpPrototype,
+      RegExpWrapper
+    );
+
+    if (UNSUPPORTED_Y) defineProperty(result, 'sticky', {
```
zloirock: It should be placed on the prototype. I'm not sure that it's 100% safe, but before users issues, we can try it.

nicolo-ribaudo: If we place it in the prototype, I'm not sure how we could check if the regexp is sticky.
Maybe I could add a `__core-js_sticky__` own property and then make `sticky` a prototype accessor for that property? I don't want to use a weakmap since it would always need to be polyfilled.

zloirock: For this case, we have https://github.com/zloirock/core-js/blob/master/packages/core-js/internals/internal-state.js

tests/tests/es.regexp.exec.js
```diff
@@ -26,3 +26,33 @@ QUnit.test('RegExp#exec capturing groups', assert => {
   // #replace, but here also #replace is buggy :(
   // assert.deepEqual(/(a?)?/.exec('x'), ['', undefined], '/(a?)?/.exec("x") returns ["", undefined]');
 });
+
+QUnit.test('RegExp#exec sticky', assert => {
+  const re = new RegExp('a', 'y');
```
zloirock: We do not polyfill it in IE8- since we can't add accessors. At least, this test should be ignored in engines without accessors.

nicolo-ribaudo: :+1:

I'm also moving it to a separate file.

nicolo-ribaudo:
I have tested this PR in modern Chrome (full sticky support), Firefox 11 (broken sticky support) and IE11 (no sticky support). Note that in Firefox 11 it is not possible to polyfill the .sticky accessor, so I had to disable some tests there.

packages/core-js-compat/src/data.js
```diff
@@ -787,6 +787,13 @@ module.exports = {
     firefox: '37',
     safari: '9.0',
   },
+  'es.regexp.sticky': {
+    chrome: '49',
+    edge: '13',
+    firefox: '3',
+    opera: '36',
+    safari: '10.0',
```
nicolo-ribaudo: I got this data from MDN. Note that on Firefox 3 the `y` flags doesn't correctly work with regexp methods, but this compat entry is only about the `.sticky` accessor.

packages/core-js/internals/fix-regexp-well-known-symbol-logic.js
```diff
@@ -20,6 +19,12 @@ var REPLACE_SUPPORTS_NAMED_GROUPS = !fails(function () {
   return ''.replace(re, '$<a>') !== '7';
 });
 
+// IE <= 11 replaces $0 with the whole match, as if it was $&
+// https://stackoverflow.com/questions/6024666/getting-ie-to-replace-a-regex-with-the-literal-string-0
+var REPLACE_KEEPS_$0 = (function () {
```
nicolo-ribaudo: If you prefer I can move this fix to a separate PR. I fixed it here just because I saw the tests failing on IE11 and it is related to RegExps.

packages/core-js-compat/src/data.js
zloirock: We don't need data for modern `opera` - it's generated from `chrome`.

packages/core-js/internals/fix-regexp-well-known-symbol-logic.js
zloirock: Nope, it's fine 👍

tests/compat/tests.js
```diff
@@ -710,6 +710,13 @@ GLOBAL.tests = {
   'es.regexp.flags': function () {
     return /./g.flags === 'g';
   },
+  'es.regexp.sticky': function () {
+    try {
+      return new RegExp('a', 'y').sticky === true;
+    } catch (e) {
+      return false;
+    }
```
zloirock: We don't need `try / catch` here - we have it in tests runner.

tests/compat/tests.js
```diff
@@ -710,6 +710,13 @@ GLOBAL.tests = {
   'es.regexp.flags': function () {
```
zloirock: Seems we should also update this and `RegExp` constructor tests and results.

packages/core-js/modules/es.string.split.js
```diff
@@ -99,24 +95,22 @@ require('../internals/fix-regexp-well-known-symbol-logic')(
         var flags = (rx.ignoreCase ? 'i' : '') +
                     (rx.multiline ? 'm' : '') +
                     (rx.unicode ? 'u' : '') +
-                    (SUPPORTS_Y ? 'y' : 'g');
```
zloirock: !

packages/core-js/internals/regexp-exec.js
```diff
@@ -18,24 +19,41 @@ var UPDATES_LAST_INDEX_WRONG = (function () {
...
 if (PATCH) {
   patchedExec = function exec(str) {
     var re = this;
     var lastIndex, reCopy, match, i;
+    var sticky = UNSUPPORTED_Y && re.sticky;
+
+    if (sticky) {
+      var flags = (re.ignoreCase ? 'i' : '') +
+                  (re.multiline ? 'm' : '') +
+                  (re.unicode ? 'u' : '') +
```
zloirock: Seems also here.

nicolo-ribaudo: I don't understand how this can throw: `re.unicode` will always be false in old envs (e.g. IE8)

packages/core-js/modules/es.string.split.js
```diff
@@ -3,19 +3,15 @@
...
 var arrayPush = [].push;
 var min = Math.min;
 var MAX_UINT32 = 0xffffffff;
 
-// babel-minify transpiles RegExp('x', 'y') -> /x/y and it causes SyntaxError
-var SUPPORTS_Y = !fails(function () { return !RegExp(MAX_UINT32, 'y'); });
-
 // @@split logic
 require('../internals/fix-regexp-well-known-symbol-logic')(
   'split',
```
zloirock: In `internalSplit`, we have possible `RegExp` call with `y` flag.

packages/core-js/internals/regexp-sticky-helpers.js
```diff
@@ -0,0 +1,42 @@
...
+  // https://bugzilla.mozilla.org/show_bug.cgi?id=773687
+  var re = RE('^r', 'gy');
+  re.lastIndex = 2;
+  return re.exec('str') != null;
+});
+
+exports.createStickyRegExp = function (re, otherFlags) {
+  var C = speciesConstructor(re, RegExp);
+
+  if (C !== RegExp) return new C(re, otherFlags + 'y');
```
zloirock: It will cause calling `RegExp` with `y` flag in tests like [this](https://github.com/zloirock/core-js/blob/698daaadbec4aab0a72d494382a4f6125b1aed08/tests/tests/es.string.split.js#L741).

ThaNarie:
I realise this issue has been stale for a while, but I would just like to add my findings here.

We've been using the fluent library in our project, which heavily relies on the sticky flag, and uses the lastIndex to move the cursor through the string while using multiple regexps with test and exec.
When merging this PR in a core-js fork we used in our project, we ran into 2 major issues, which I think still need to be addressed here:

1) `regexp.exec` wasn't included by babel-preset-env because it still has `es.regexp.exec` as implemented in `ie: '9',`: https://github.com/zloirock/core-js/blob/3699c84d07fd63622e8fbbdd836e5cf4fd4e8a79/packages/core-js-compat/src/data.js#L792
Removing this line properly polyfills the `exec` for IE.

2) The sticky flag should also modify the behaviour of `regexp.test` by moving the `lastIndex` property the same way as `exec` does. But since there is no `test` polyfill included in core-js at all, this behaviour doesn't work in IE.

A quick fix for this is to patch `regexp.test` with `return !!re.exec(str);`.

Repro that fails in IE:
```
var re = new RegExp('\\s+', 'y')
re.test('  aa');
console.log(re.lastIndex); // should be 2, but is 0 in IE with current branch
```

zloirock:
@ThaNarie since that PR is suspended, feel free to make an alternative version if it's interesting for you.

nicolo-ribaudo:
@ThaNarie I don't have much time to pick this up again, buy if you want you could just take this PR and modify it if it's easier!

HankMcCoy:
My team has been running into the same issues w/ Fluent that @ThaNarie mentioned, so we are planning on taking a shot at getting an updated version of this PR up.

We haven't contributed to core-js before, so we'll dig into the [CONTRIBUTING](https://github.com/zloirock/core-js/blob/master/CONTRIBUTING.md) doc and see what we can figure out!

zloirock:
@HankMcCoy if you will have questions, ask me -)

cvle:
@HankMcCoy I will also dedicate some time in the coming weeks. It would be awesome if we could collaborate.

PR #510, opened 2019-03-21
fix: fixes code clarity with isThenable and ESLint issue (closes #509)

### LINKED ISSUE #509: Assignment instead of conditional expression
I noticed that in conditionals there are assignments and not expressions:

```sh
rg "then = isThenable" node_modules

node_modules/core-js/client/core.js
6685:          } else if (then = isThenable(result)) {
6752:    if (then = isThenable(value)) {

node_modules/core-js/client/library.js
5846:          } else if (then = isThenable(result)) {
5913:    if (then = isThenable(value)) {

node_modules/core-js/client/shim.js
6583:          } else if (then = isThenable(result)) {
6650:    if (then = isThenable(value)) {

node_modules/core-js/modules/es6.promise.js
83:          } else if (then = isThenable(result)) {
150:    if (then = isThenable(value)) {

node_modules/core-js-pure/modules/es.promise.js
105:          } else if (then = isThenable(result)) {
185:    var then = isThenable(value);

node_modules/core-js/library/modules/es6.promise.js
83:          } else if (then = isThenable(result)) {
```

niftylettuce:


packages/core-js/modules/es.promise.js
```diff
@@ -100,9 +100,11 @@ var notify = function (promise, state, isReject) {
               exited = true;
             }
           }
+          var isResultThenable = isThenable(result);
```
zloirock: It's an observable change, so it should be done after the first `if`.

zloirock:
Feel free to add this rule to `core-js` eslint config.

packages/core-js/modules/es.promise.js
niftylettuce: Fixed

niftylettuce:
I've made the requested change per your comment, thank you @zloirock!

zloirock:
So... Could you add this rule to the `eslint` config?

niftylettuce:
There's no need to add to the rule if my PR is merged?

niftylettuce:
Oh I see what you mean, yes one sec

niftylettuce:
Done @zloirock

zloirock:
After linting, we have 26 (16 without duplicates) related warnings.

zloirock:
@niftylettuce do you wanna fix the rest issues?

niftylettuce:
I don't have time yet 😦

PR #511, opened 2019-03-21
Fix duplicated assignment to module.exports

cmfcmf:
While looking through the code, I found some duplicated `module.exports` assignments, which were probably copy-paste errors. This PR replaces them all :spaghetti:


zloirock:
Thanks -)

Issue #536, opened 2019-04-26
RegExp.toString gives incorrect result without RegExp.flags


bz2:
The `RegExp.toString` polyfill creates invalid patterns in the form `/a/undefined` unless the browser either supports `RegExp.flags` natively or that is also polyfilled. This is an issue when trying to use targetted imports as created by the `@babel/preset-env` setting `useBuiltIns: 'usage'`.

From the compat data, this is a problem for browsers like IE 11 and Edge, as Chrome/Firefox and so on have had native support for a while:
https://github.com/zloirock/core-js/blob/v3.0.1/packages/core-js-compat/src/data.js#L786

Tracking this down took a while, so I created minimal-ish project that reproduces the problem for demonstration:
https://github.com/zegami/test_core_regexp

bz2:
As a workaround, if a file depend on `toString()` for `RegExp` can manually include the core-js import for regexp/flags too:

```js
// TODO: Remove when zloirock/core-js#536 is fixed upstream.
import 'core-js/es/regexp/flags';
```

This works even if the regexp/to-string polyfill is autocreated by `@babel/preset-env`.

PR #537, opened 2019-04-26
Give correct RegExp flags in toString

bz2:
When the flags property is not available on `RegExp`, the internals `regexp-flags` implementation should be used regardless of `DESCRIPTORS`.

Fixes: zloirock/core-js#536

I'm not *certain* this is the right fix, but the value of `DESCRIPTORS` is `true` on IE 11 and the internals implementation would be fine - and doesn't seem to depend on `defineProperty`.

Also not sure how to include a focused test, but the linked issue has a project that reproduces the problem.

### DIFF
--- a/packages/core-js/modules/es.regexp.to-string.js
+++ b/packages/core-js/modules/es.regexp.to-string.js
@@ -2,9 +2,9 @@
 var anObject = require('../internals/an-object');
 var fails = require('../internals/fails');
 var flags = require('../internals/regexp-flags');
-var DESCRIPTORS = require('../internals/descriptors');
 var TO_STRING = 'toString';
 var nativeToString = /./[TO_STRING];
+var RegExpPrototype = RegExp.prototype;
 
 var NOT_GENERIC = fails(function () { return nativeToString.call({ source: 'a', flags: 'b' }) != '/a/b'; });
 // FF44- RegExp#toString has a wrong name
@@ -15,7 +15,8 @@ var INCORRECT_NAME = nativeToString.name != TO_STRING;
 if (NOT_GENERIC || INCORRECT_NAME) {
   require('../internals/redefine')(RegExp.prototype, TO_STRING, function toString() {
     var R = anObject(this);
+    var rf = R.flags;
     return '/'.concat(R.source, '/',
-      'flags' in R ? R.flags : !DESCRIPTORS && R instanceof RegExp ? flags.call(R) : undefined);
+      rf == null && R instanceof RegExp && !('flags' in RegExpPrototype) ? flags.call(R) : rf);
   }, { unsafe: true });
 }
--- a/tests/tests/es.regexp.to-string.js
+++ b/tests/tests/es.regexp.to-string.js
@@ -22,6 +22,11 @@ QUnit.test('RegExp#toString', assert => {
     flags: 'bar',
   }), '/foo/bar');
   assert.same(toString.call({}), '/undefined/undefined');
+  assert.same(toString.call(new Proxy(/pattern/, {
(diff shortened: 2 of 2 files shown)


zloirock:
I think that there should be something like
```js
var R = anObject(this);
var p = String(R.source);
var rf = R.flags;
var f = String(rf === undefined && R instanceof RegExp && !('flags' in RegExpPrototype) ? flags.call(R) : rf);
return '/' + p + '/' + f;
```

bz2:
@zloirock Thanks for looking over this. I've read these parts of the spec now so have a slightly better understanding.

In your suggested implementation, I'm not sure I understand the purpose of checking `RegExpPrototype`. Any conforming flags getter cannot return `undefined`, only raise a `TypeError` or return a string. So, if `rf` is not a string but `R` is a `RegExp` instance the code should always use the internals flags implementation instead?

https://www.ecma-international.org/ecma-262/6.0/#sec-get-regexp.prototype.flags
https://www.ecma-international.org/ecma-262/6.0/#sec-regexp.prototype.tostring

On the tests, I have a plan that would sort of work, but is likely too big a change for this branch. Main issue is the tests are always run in a browser context that includes the whole polyfill bundle:

`tests/tests.html`:
```html
<script src='../packages/core-js-bundle/index.js'></script>
<!-- ... -->
<script src='./bundles/tests.js'></script>
```

It would be possible to change `.webpack.config.js` for instance, to have an entry per test file, rather than one for all tests:
```js
  tests: {
    entry: './tests/tests/index.js',
    output: { filename: 'tests.js' },
  },
```
That would then make it possible to test each file with only its own polyfill rather than all of core-js. That still doesn't help if the tests are run on a modern chrome where the feature is implemented anyway, but is heading in a sensible direction?

zloirock:
> In your suggested implementation, I'm not sure I understand the purpose of checking `RegExpPrototype`...

It can be called on non-regex or a `RegExp` subclass instances. As you can see in the spec, those methods are generics. `flags` property can be redefined. Internal implementation can't be used always since it could have side effects in the same cases. So in my code above you can see an example of correct fallback for engines without `RegExp.prototype.flags`.

zloirock:
You have correctly identified a problem of testing. However, the problem is deeper than you think - for example, a big part of modules are not independent and should work in different ways in different combinations and those combinations defined in many different places like CommonJS entry point, `@babel/preset-env` definitions, etc.

bz2:
@zloirock Have poked around and can't come up with a scenario that doesn't work correctly with the `in` instance check and `instanceof`. Plain objects use `.flags`, only things with `RegExp` in the prototype chain can get the internals implementation, and it doesn't really seem possible to subclass `RegExp` anyway? If you can give an example test case that fails, I'll adapt the logic.

zloirock:
@bz2 it's not a problem to write an example with usage of `Proxy` where we can redefine `in` behavior on `R`. However, it's also required for consistency with other cases like [this](https://github.com/zloirock/core-js/blob/1e67304924030e8f251802f4775f2d8517ab3aef/packages/core-js/modules/esnext.string.match-all.js#L62).

bz2:
@zloirock Have updated with basic proxy test (that would have failed on non-flags browsers only) and that spelling. Does seem like the same form should be used in `match-all` and here, but that code is currently not caring about being passed non-RegExp objects?

zloirock:
@bz2 since we can't polyfill `Proxy` and early `Proxy` implementations have inconsistent API, I don't think that adding a test with `Proxy` makes sense. Sure, `match-all` should be updated, but in the scope of #516. I think I will do it myself.

PR #539, opened 2019-04-26
Fix broken getOwnPropertySymbols in Mobile Chrome 38 and 39

nicketson:
This is a workaround for this chrome issue
https://bugs.chromium.org/p/v8/issues/detail?id=3443

Without this fix, the object spread operator is broken for these
versions of chrome when transpiling and shimming with babel and
core-js@3.
See https://github.com/babel/babel/issues/8721

### DIFF
--- a/packages/core-js/internals/object-get-own-property-symbols-external.js
+++ b/packages/core-js/internals/object-get-own-property-symbols-external.js
@@ -1,6 +1,0 @@
-var toIndexedObject = require('../internals/to-indexed-object');
-var nativeGetOwnPropertySymbols = require('../internals/object-get-own-property-symbols').f;
-
-module.exports.f = function getOwnPropertySymbols(it) {
-  return nativeGetOwnPropertySymbols(toIndexedObject(it));
-};
--- a/packages/core-js/internals/object-get-own-property-symbols-external.js
+++ b/packages/core-js/internals/object-get-own-property-symbols-external.js
@@ -1,0 +1,6 @@
+var toIndexedObject = require('../internals/to-indexed-object');
+var nativeGetOwnPropertySymbols = require('../internals/object-get-own-property-symbols').f;
+
+module.exports.f = function getOwnPropertySymbols(it) {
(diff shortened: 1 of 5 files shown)


zloirock:
Could you also update `tests/tests/es.symbol.js`, `tests/pure/es.symbol.js`, `tests/compat/tests.js`?

packages/core-js/internals/object-get-own-property-symbols-external.js
```diff
@@ -0,0 +1,6 @@
+var toIndexedObject = require('../internals/to-indexed-object');
+var nativeGetOwnPropertySymbols = require('../internals/object-get-own-property-symbols').f;
+
+module.exports.f = function getOwnPropertySymbols(it) {
+  return nativeGetOwnPropertySymbols(toIndexedObject(it));
+};
```
zloirock: I think that it can be moved to `es.symbol` since it's used only here.

zloirock: `toIndexedObject` required only for old IE where we can access to string chars by index, those engines haven't native `Object.getOwnPropertySymbols`, so here should be used `toObject`.

nicketson: Understood. I just got the test suite working. I'll commit some updates in the next day or two.

nicketson: Moved `packages/core-js/internals/object-get-own-property-symbols-external.js` into `es.symbol` and switched `toIndexedObject` to `toObject`. Also fixed failing core-js-pure unit tests while addressing `packages/core-js/internals/object-get-own-property-symbols-external.js`

nicketson:
Does this look good now? Do you want me to rebase it?

zloirock:
Yep, LGTM. Only one (mainly stylistic) moment - I'd prefer to have only one export of `getOwnPropertySymbols` in `es.symbol`. It's not critical and I can change it later.

PR #576, opened 2019-06-18
respect DISABLE_OPENCOLLECTIVE env variable

slang800:
The main [opencollective-postinstall](https://www.npmjs.com/package/opencollective-postinstall) package uses the `DISABLE_OPENCOLLECTIVE` environment variable to determine whether or not the ad should be shown. The webpack-cli package [does this too](https://github.com/webpack/webpack-cli/blob/08edf5766afc3201689eeaf56cfaaad567a22633/bin/opencollective.js#L40).

This PR adds support for hiding the open collective ad when `DISABLE_OPENCOLLECTIVE` is set to true.

### DIFF
--- a/packages/core-js/scripts/postinstall.js
+++ b/packages/core-js/scripts/postinstall.js
@@ -3,6 +3,7 @@ var env = process.env;
 var ADBLOCK = is(env.ADBLOCK);
 var CI = is(env.CI);
 var COLOR = is(env.npm_config_color);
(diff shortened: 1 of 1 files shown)


zloirock:
I already wrote that I'm not sure that it's a good idea since it's not only Open Collective. However, since it's not the first issue about this flag - OK. Please, place it near `ADBLOCK` env variable which made for the same.

PR #585, opened 2019-06-27
Smiley correction

ajmeese7:
All your smiley faces were cyclops, but now they are happy two-eyed smiles =)

Sorry if they were intended to be one-eyed for a certain reason. I don't mean to offend anyone, I just assumed this was a mistake and attempted to rectify it. I *think* these are supposed to be emoji faces but I might be way off base here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 **If you looking documentation for obsolete `core-js@2`, please, check [this branch](https://github.com/zloirock/core-js/tree/v2).**
 
-## As advertising: the author is looking for a good job -)
+## As advertising: the author is looking for a good job =)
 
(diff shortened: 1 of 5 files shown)


trdarr:
[this is important. this means something.](https://www.youtube.com/watch?v=UQG5rsCtU2M)

ErikHumphrey:
Can't fix this one, though:
image

It's likely related to this: [Why do Russians use ")" as a smiley instead of ":)"?](https://www.quora.com/Why-do-Russians-use-as-a-smiley-instead-of)

Thanks, DotA.

zloirock:
@ErikHumphrey is right.

I'd prefer to leave it as is. Thanks for the PR.

PR #590, opened 2019-07-04
[core-js-compat] Filter available modules without using an intermediate set

nicolo-ribaudo:
Ref: https://github.com/babel/babel/pull/10158#issuecomment-508285566

### DIFF
--- a/packages/core-js-compat/package.json
+++ b/packages/core-js-compat/package.json
@@ -12,7 +12,6 @@
     "semver": "^6.1.3"
   },
   "devDependencies": {
-    "core-js-pure": "3.1.4",
     "detective": "^5.2.0",
     "electron-to-chromium": "^1.3.180",
     "glob": "^7.1.4"
--- a/packages/core-js-compat/package.json
+++ b/packages/core-js-compat/package.json
@@ -9,10 +9,10 @@
   "main": "index.js",
   "dependencies": {
     "browserslist": "^4.6.3",
-    "core-js-pure": "3.1.4",
     "semver": "^6.1.3"
   },
   "devDependencies": {
+    "core-js-pure": "3.1.4",
     "detective": "^5.2.0",
     "electron-to-chromium": "^1.3.180",
     "glob": "^7.1.4"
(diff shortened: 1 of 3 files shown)


packages/core-js-compat/package.json
```diff
@@ -9,10 +9,10 @@
   "main": "index.js",
   "dependencies": {
     "browserslist": "^4.6.3",
-    "core-js-pure": "3.1.4",
     "semver": "^6.1.3"
   },
   "devDependencies": {
+    "core-js-pure": "3.1.4",
```
nicolo-ribaudo: This is still used by `packages/core-js-compat/src/build-entries.js` (which isn't included in the published package, so it can be a dev dependency).

zloirock: Since it's a similar case like in `index`, it would be better also replace it.

packages/core-js-compat/src/build-entries.js
```diff
@@ -19,7 +18,7 @@ function getModulesForEntryPoint(entry) {
     const relative = resolve(dir, dependency);
     result.push(...getModulesForEntryPoint(relative));
   }
-  return [...intersection(new Set(result), order)];
+  return result.sort((a, b) => order.indexOf(a) > order.indexOf(b) ? 1 : -1);
```
zloirock: It's slow -/

nicolo-ribaudo: Oh I tought that you meant to remove `intersection` like this. Would it be ok to leave it there as it was before?

zloirock: Something like
```js
const resultSet = new Set(result);
return order.filter(it => resultSet.has(it));
```
could be better.

zloirock: `O(n)` vs `O(n^2 log n)` if I understood correctly.

nicolo-ribaudo: Yeah good idea.

nicolo-ribaudo:
Thanks!

PR #593, opened 2019-07-11
update core-js-compat for Firefox 67

henryqdineen:
Looks like `es.string.match-all` is supported in Firefox 67. MDN also confirms this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll#Browser_compatibility

Thanks!


zloirock:
Thanks.

PR #594, opened 2019-07-11
update core-js-compat for Firefox 69

henryqdineen:
PromiseRejection has been enabled in Firefox 69 (https://bugzilla.mozilla.org/show_bug.cgi?id=1525554). This is currently in Nightly. I wasn't sure if we should wait until it is past beta to update `core-js-compat`. We plan to use `core-js-compat` as part of our polyfilling solution so it would be good to have this data in there as soon as we start to get traffic from Firefox 69.

Thanks!


zloirock:
The problem that it's is currently Nightly, so I'm not sure that all those features will be added to the stable FF69. Let's wait for the FF Beta.

PR #597, opened 2019-07-17
Fix repeatedly showing postinstall to once a day

remy:
This uses a temp file and checks the timestamp (useful for local development) and it means that the request for support is only shown on that particular machine once a day.

Ref #548 

Hopefully this helps calm people down!

### DIFF
--- a/packages/core-js/scripts/postinstall.js
+++ b/packages/core-js/scripts/postinstall.js
@@ -9,10 +9,12 @@ var SILENT = !!~['silent', 'error', 'warn'].indexOf(env.npm_config_loglevel);
 var tmpdir = require('os').tmpdir();
 var fs = require('fs');
 
+var minute = 60 * 1000;
+
 var SEEN = false;
 try {
   var delta = Date.now() - fs.statSync(tmpdir + '/core-js-banner').mtime;
-  SEEN = delta < 24 * 60 * 60 * 1000;
+  SEEN = delta < 5 * minute;
 } catch (e) {}
 
 function is(it) {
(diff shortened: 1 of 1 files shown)


zloirock:
Thanks. One moment - I'm ready to limit it to one message per `npm install`. I have no ideas about how to get this info, so we could use, for example, 5 minutes. Also, `try / catch` block on `writeFileSync` could be useful.

balibebas:
@remy, loved your HTML5 book. Someone totally stole the copy I purchased a few years back it was that good.

packages/core-js/scripts/postinstall.js
```diff
@@ -15,10 +24,12 @@ function log(it) {
   console.log(COLOR ? it : it.replace(/\u001B\[\d+m/g, ''));
```
JeremyRH: Could `console.info` be used instead of `console.log`? `console.log` is not properly silenced when the `--loglevel=warn` argument is used.

remy:
Sorry, super slack on reply to this thread (though now it might be that npm will be shooting us down anyway).

I added the try/catch on write and I've changed the check to be on a 5 minute timeout as you requested.

zloirock:
Thanks. I already reworked it by a little more complex logic. About NPM - [seems new rules are not a problem](https://github.com/zloirock/core-js/issues/635#issuecomment-526792928).

onexdata:
I was hired to discover and fix the reason node build logs just doubled in size.

It's because someone added something that uses core-js, and each sub-package that uses it also shows an add saying zloirock is looking for a good job.

While I understand we all need to eat, 50+ ads spamming the same thing is apparently a big enough issue to pay someone to get rid of.

I see this has been a problem for people for a while... Might want to fix this before "zloirock is looking for a good job" becomes a meme or synonymous with spam.  That would probably not be the outcome you're looking for.

azah:
> It's because someone added something that uses core-js, and each sub-package that uses it also shows an add saying zloirock is looking for a good job.

This is still occurring in build logs.

lifenstein:
It's possible to add one of many environment variables to hide the ads: 

https://github.com/zloirock/core-js/blob/master/packages/core-js/postinstall.js#L31

azah:
So instead of outright removing the spam, making people dig into changelogs/github issues to figure out that it's even possible to disable it is better?

lifenstein:
@azah 
1. There's a bit of a backstory behind the ad: https://github.com/zloirock/core-js/issues/548#issuecomment-510684777
1. Since the sole package maintainer seems to be in jail for 18 months ( https://github.com/zloirock/core-js/issues/767#issuecomment-599031389 ), I think it is safe to assume there won't be any releases anytime soon. 
1. It's an open-source repo (MIT license) and anyone is free to fork it.
1. I never said or even implied that what I posted was "better", just a workaround that I found by digging around. I have nothing to do with the package; just pointed out that there are options available, if one bothers to look.

PR #606, opened 2019-07-24
Check for backwards slashes in module regex

### LINKED ISSUE #605: build-entries script for core-js-compat does not contain modules on Windows
The `build-entries` scripts gives the following `entries.json` on Windows:

```
{
  "core-js": [],
  "core-js/es/array-buffer/constructor": [],
  ...
  "core-js/modules/web.url.to-json": []
}
```

All the arrays are empty, because a regex that is looking for modules in `getModulesForEntryPoint` is assuming forward slashes.

flut1:
Check for both backwards and forwards slashes so the regex matches paths found by windows as well. Fixes #605


zloirock:
Thanks.

PR #607, opened 2019-07-25
fix: use correct prototype when feature detecting String.prototype.trimEnd and trimStart

JakeChampion:


zloirock:
Ooops... Thanks!

JakeChampion:
You're welcome :-) It's always nice to help out on another polyfill project :-)

PR #608, opened 2019-07-25
Avoid conflicts between Promise from core-js and Promise from zone.js

marcospont:
When loaded, zone.js will patch global Promise.
When loaded, core-js will run a detection routine to determine if it needs to patch global Promise.

If core-js is loaded before zone.js, everything works fine.
If core-js is loaded after zone.js and the order of these events cannot be changed, core-js will fail to detect a patched global Promise, causing its detection code to patch it again.

This pull request aims to improve the detection routine to avoid conflicts between core-js Promise polyfill and zone.js Promise polyfill.

### DIFF
--- a/packages/core-js/modules/es.promise.js
+++ b/packages/core-js/modules/es.promise.js
@@ -59,6 +59,7 @@ var FORCED = isForced(PROMISE, function () {
   // unhandled rejections tracking support, NodeJS Promise without it fails @@species test
   return !((IS_NODE || typeof PromiseRejectionEvent == 'function')
     && (!IS_PURE || promise['finally'])
+    && typeof promise.constructor === 'object'
     && promise.then(empty) instanceof FakePromise
     // v8 6.6 (Node 10 and Chrome 66) have a bug with resolving custom thenables
     // https://bugs.chromium.org/p/chromium/issues/detail?id=830565


zloirock:
Sorry, I'm not sure how this PR could help in this situation. `core-js` should be loaded before any other scripts since it's a polyfill. `zone.js` `Promise` is not complete and can't and shouldn't pass `core-js` feature detection.

marcospont:
I know that core-js should always be loaded first since it is a polyfill. However, in this scenario, the load order cannot be changed.

Application A is written in Angular and loads zone.js. Application B has the ability to load other applications and embed them. One of these applications is Application B, which uses AngularJS transpiled with Babel and core-js.

Assuming that Application A cannot be changed to load core-js before zone.js, the issue cannot be avoided.

In this PR, the code change avoids the conflict by testing if promise.constructor is an object just like it was assigned while creating "FakePromise". In the scenario where the conflict happens, promise.constructor will be a function just like defined by the patched Promise from zone.js.

If zone.js is loaded before core-js, calling "promise.then(empty)" will throw an error because the "then" method implemented in zone.js tries to call "new this.constructor(null)". Since the constructor is not a function, it cannot be called using "new".

zloirock:
So, maybe, you should use [this](https://github.com/zloirock/core-js#configurable-level-of-aggressiveness)?

zloirock:
@luisrudge 
- `zone.js` should be used after `core-js`. Otherwise, you will get other problems as well.
- `zone.js` promises do not match the spec, so they don't pass `core-js` feature detection. It should be fixed on `zone.js` side.
- This PR definitely is an incorrect fix for this problem since it just breaks `core-js` feature detection. Since nothing was fixed, I'm closing it.
- It's a bad idea to add global polyfills to libraries. Try to use a `core-js` version without global pollution, see `core-js-pure` / `@babel/runtime-corejs3`.

PR #680, opened 2019-10-29
Fix an error when using URLSearchParams with fetch

huww98:
Fix the following error occurred in WeChat Windows client (based on Chrome 53 from UA), when passing an `URLSearchParams` as body and not passing headers.

message: "Failed to construct 'Headers': No matching constructor signature."
stack: TypeError: Failed to construct 'Headers': No matching constructor signature.
 at TypeError (native)
 at fetch (webpack-internal:///./node_modules/core-js/modules/web.url-search-params.js:328:23)
 at _callee$ (webpack-internal:///./src/api/internal.ts:180:20)
 at tryCatch (webpack-internal:///./node_modules/regenerator-runtime/runtime.js:45:40)
...


zloirock:
Hmm... Interesting, thanks.

zloirock:
It seems to need to check it with some other early `fetch` implementations.

PR #681, opened 2019-10-29
[WIP] Well‑formed `JSON.stringify`

zloirock:


packages/core-js/modules/es.json.stringify.js
```diff
@@ -0,0 +1,43 @@
...
+    || $stringify('\uDEAD') !== '"\\udead"';
+});
+
+if ($stringify) {
+  // https://github.com/tc39/proposal-well-formed-stringify
+  $({ target: 'JSON', stat: true, forced: FORCED }, {
+    // eslint-disable-next-line no-unused-vars
+    stringify: function stringify(it, replacer, space) {
+      var result = $stringify.apply(null, arguments);
+      return typeof result == 'string' && reg.test(result) ? fix(result) : result;
```
ExE-Boss: Does&nbsp;it&nbsp;even make&nbsp;sense to&nbsp;iterate&nbsp;over the&nbsp;string&nbsp;twice? 🤔

First&nbsp;time it’s&nbsp;done by&nbsp;`reg.test(result)`, and&nbsp;the&nbsp;second time it’s&nbsp;done by&nbsp;`fix(result)`.

zloirock: @ExE-Boss that prevents the creation of a new string in `fix` and slower iteration. It seems to need to write a little more complex `reg` regex to prevent any false positives. Maybe makes sense do it once in `.replace`, however, I'm not sure.

zloirock: Ok, let's simplify it with `.replace`.

packages/core-js/modules/es.json.stringify.js
```diff
@@ -3,27 +3,16 @@ var getBuiltIn = require('../internals/get-built-in');
...
-    if (re.test(point)) {
-      prev = string.charAt(i - 1);
-      next = string.charAt(i + 1);
-      if ((low.test(point) && !hi.test(next)) || (hi.test(point) && !low.test(prev))) {
-        result += '\\u' + point.charCodeAt(0).toString(16);
-        continue;
-      }
-    } result += point;
-  } return result;
+var fix = function (match, offset, string) {
```
ExE-Boss:
```suggestion
var fix = function (match, prev, char, next, offset, string) {
```

packages/core-js/modules/es.json.stringify.js
```diff
@@ -3,27 +3,16 @@ var getBuiltIn = require('../internals/get-built-in');
 var fails = require('../internals/fails');
 
 var $stringify = getBuiltIn('JSON', 'stringify');
-var reg = /[\uD800-\uDFFF]/g;
-var re = /^[\uD800-\uDFFF]$/;
+var re = /[\uD800-\uDFFF]/g;
```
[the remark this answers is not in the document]
zloirock: It will cause [this problem](https://github.com/ungap/json/issues/2#issuecomment-547512527).

packages/core-js/modules/es.json.stringify.js
```diff
@@ -3,27 +3,16 @@ var getBuiltIn = require('../internals/get-built-in');
...
-      next = string.charAt(i + 1);
-      if ((low.test(point) && !hi.test(next)) || (hi.test(point) && !low.test(prev))) {
-        result += '\\u' + point.charCodeAt(0).toString(16);
-        continue;
-      }
-    } result += point;
-  } return result;
+var fix = function (match, offset, string) {
+  var prev = string.charAt(offset - 1);
+  var next = string.charAt(offset + 1);
```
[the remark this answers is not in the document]
zloirock: That returns an empty string, it's an acceptable value for the next test.

packages/core-js/modules/es.json.stringify.js
zloirock: See above.

packages/core-js/modules/es.json.stringify.js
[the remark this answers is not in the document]
ExE-Boss: I wasn’t aware of that.

I would just expect it to throw like it does in **Java**, **C#** or any other sane programming language.

PR #687, opened 2019-11-03
Fix Math.signbit semantics

chicoxyzzy:
According [spec draft](https://tc39.es/proposal-Math.signbit/Math.signbit.html#spec), Math.signbit returns `true` only for negative values and `Math.signbit(NaN)` should return `false`


chicoxyzzy:
Oops forgot to fix test. I'll update my PR

zloirock:
Hmm...

zloirock:
IIRC, this polyfill in the `core-js` from the time when the spec was not ready. So seems I messed up something.

zloirock:
Thanks.

Issue #582, opened 2019-06-24
Implement postinstall error handling in a more cross-platform way


ilkka:
Currently the [postinstall script](https://github.com/zloirock/core-js/blob/29c40031624a5d90fdb95ab414e5f690879d3eb7/packages/core-js/package.json#L49) eats errors with a `|| echo 'ignore'` construct. This does not work in PowerShell, and therefore the package is not installable by anybody foolish or brash enough to configure their npm `script-shell` to be PowerShell. Surely the postinstall script itself could be made to ignore any typical errors by itself? Can somebody shed some light on why it even has that `echo` at the end?

I realize this may be a bit niche but it'd still be a nice little enhancement not to have to rely on specific shell features that are not universal.

zloirock:
Because it's more cross-platform than alternatives like
```js
node scripts/postinstall || true
```
or
```js
node -e \"try { require('./scripts/postinstall') } catch (_) {}\"
```
Sure it's not perfect, so feel free to propose better sollution for preventing possible errors on `postinstall`.

ilkka:
Alright, I see! Do you happen to remember if there were any specific error conditions that this solution was created to solve? The script seems so simple that one would think that it would only fail if the environment was restricted in a very specific way, like having `console.log` throw if called.

zloirock:
It's many different errors in many different environments. Some of them you could find in #551 or comments of #548.

dmwyatt:
Yeah, this broke a ton of my builds.

I'm not sure I agree this is more "cross-platform".  This change makes core-js just not work at all on one platform.

zloirock:
@dmwyatt wdym?

dmwyatt:
`||` is not supported in powershell.

zloirock:
Feel free to propose a fix.

dmwyatt:
I don't yet fully grasp the point of the recent change or the point of your postinstall script. However, I do know that this breaks the install for everyone using powershell.

My first reaction is to just get rid of your current postinstall script.

My second reaction is: it worked fine on powershell before the recent change, so revert that change.  Is the income stream from developers who sit and read their npm logs that significant?

The Powershell equivalent is something like `(node scripts/postinstall.js) -OR (echo "ignore")`.

I've used [run-script-os](https://github.com/charlesguse/run-script-os) for doing cross-platform scripts before, though I've never used it for `postinstall`.

Would you consider a PR that uses that package to solve this issue?

zloirock:
[...]

PR #690, opened 2019-11-07
fix: post installs fails with powershell

dopry:
with npm config set script-shell powershell.exe the postinstall hook
failes with the error

```
> node postinstall || echo "ignore"

At line:1 char:18
+ node postinstall || echo "ignore"
+                  ~~
The token '||' is not a valid statement separator in this version.
    + CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : InvalidEndOfLine
```

### DIFF
--- a/packages/core-js-bundle/package.json
+++ b/packages/core-js-bundle/package.json
@@ -48,6 +48,6 @@
     "shim"
   ],
   "scripts": {
-    "postinstall": "node postinstall || echo \"ignore\""
+    "postinstall": "node -e \"try { require('./scripts/postinstall'); } catch (e) { }; process.exit(0) \""
   }
 }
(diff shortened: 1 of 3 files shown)


dopry:
applies to #582

packages/core-js-bundle/package.json
```diff
@@ -48,6 +48,6 @@
     "shim"
   ],
   "scripts": {
-    "postinstall": "node postinstall || echo \"ignore\""
+    "postinstall": "node -e \"try { require('./scripts/postinstall'); } catch (e) { }; process.exit(0) \""
```
zloirock: `'./scripts/postinstall'` -> `'./postinstall'`.

Is `process.exit(0)` necessary?

zloirock: @dopry any news?

PR #692, opened 2019-11-10
Remove copyright from executable code

coreyward:
This winds up in the minified distribution everywhere this library is used instead of the usual inclusion on a page with a list of the open source software used (i.e. licenses page). It's a minor thing, but multiplied by millions and millions of instances of the notice across likely billions of requests.

### DIFF
--- a/packages/core-js/internals/shared.js
+++ b/packages/core-js/internals/shared.js
@@ -6,5 +6,4 @@ var store = require('../internals/shared-store');
 })('versions', []).push({
   version: '3.4.0',
   mode: IS_PURE ? 'pure' : 'global',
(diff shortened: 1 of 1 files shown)


zloirock:
That helps identify where used `core-js`. Thanks for the PR, but I don't think that it's a good idea.

coreyward:
Okay, so what if every library did it? We’d have a few dozen KB worth of tacky bathroom stall graffiti on practically every website. And for what value? You can already see who is using your library through the code and other signatures.

zloirock:
I don't think that most other libraries cause so many problems on debugging - at least finding the place, where, how used and how many copies. Because of the extra modularity, other signatures are the bad identifier. Even if each library will do it - it's not my problem -)

PR #694, opened 2019-11-12
Add enterprise language

phated:
@zloirock as per our email thread, here is a PR for the new Tidelift enterprise language. I also removed Tidelift from the other section because we are not a donation platform.


zloirock:
Could you move it under the next section?

phated:
@zloirock moved! I wanted to note that we've found that the further down in a README this language is, the less it converts—feel free to reach out if you have any other questions.

PR #701, opened 2019-11-21
Fix smiley in postinstall message

herschel666:


zloirock:
Hi. Thanks, but it seems a duplicate of #585.

PR #702, opened 2019-11-21
Check if startsWith is configurable

pdemro:
String.startsWith polyfill will throw an exception if the existing property is not configurable.  Other polyfills will likely throw an exception under the same scenario.  Perhaps an update like this belongs farther up the chain.

This still needs to be tested.  Wanted to start the PR process to open the discussion

### DIFF
--- a/packages/core-js/modules/es.string.starts-with.js
+++ b/packages/core-js/modules/es.string.starts-with.js
@@ -8,16 +8,18 @@ var correctIsRegExpLogic = require('../internals/correct-is-regexp-logic');
 var nativeStartsWith = ''.startsWith;
 var min = Math.min;
 
-// `String.prototype.startsWith` method
-// https://tc39.github.io/ecma262/#sec-string.prototype.startswith
-$({ target: 'String', proto: true, forced: !correctIsRegExpLogic('startsWith') }, {
-  startsWith: function startsWith(searchString /* , position = 0 */) {
-    var that = String(requireObjectCoercible(this));
-    notARegExp(searchString);
-    var index = toLength(min(arguments.length > 1 ? arguments[1] : undefined, that.length));
-    var search = String(searchString);
-    return nativeStartsWith
-      ? nativeStartsWith.call(that, search, index)
-      : that.slice(index, index + search.length) === search;
-  }
-});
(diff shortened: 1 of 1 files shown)


packages/core-js/modules/es.string.starts-with.js
```diff
@@ -8,16 +8,18 @@ var correctIsRegExpLogic = require('../internals/correct-is-regexp-logic');
...
-    var that = String(requireObjectCoercible(this));
-    notARegExp(searchString);
-    var index = toLength(min(arguments.length > 1 ? arguments[1] : undefined, that.length));
-    var search = String(searchString);
-    return nativeStartsWith
-      ? nativeStartsWith.call(that, search, index)
-      : that.slice(index, index + search.length) === search;
-  }
-});
+if(Object.getOwnPropertyDescriptor(String.prototype, "startsWith").configurable) {
```
pdemro: If the property is not configurable, do nothing since it is going to fail anyway with an exception

JakeChampion: When would it not be configurable?

zloirock:
It could be non-configurable only with incorrect polyfill, like a polyfill from MDN. It would be better just to remove incorrect polyfill.

pdemro:
> 
> 
> It could be non-configurable only with incorrect polyfill, like a polyfill from MDN. It would be better just to remove incorrect polyfill.

I agree 100%.  Here is my scenario (IE11)

1.  My code installs/imports dependency X
2. Dependency X installs/imports core-js/modules/es.string.starts-with
3. Embed my code into cloud service which already created a polyfill for startsWith (that is non-configurable for some reason)

In this case the cloud service is Microsoft 365 SharePoint (using the SharePoint Framework) and "dependency x" is the botframework-webchat es5 bundle https://github.com/microsoft/BotFramework-WebChat/tree/master/samples/01.b.getting-started-es5-bundle

zloirock:
Maybe it would be better to start from fixing this issue on MDN and related sources?

zloirock:
I'm not a fan of adding workarounds for other polyfills bugs - that means incorrect usage of polyfills. However, it's not the first mention of this issue and if it would be fixed in the source of the problem, I'll think about adding a workaround.

zloirock:
Still not fixed on MDN side:
image
image

PR #717, opened 2019-12-04
Add test to ensure Reflect.ownKeys works with a Symbol polyfill

JakeChampion:


zloirock:
Thanks for the PR, but I'm not sure that it's required. Symbols can't be completely implemented only on the polyfills layer, so for `core-js` `Symbol` polyfill is no difference, is this `Symbol()` or `Object(Symbol())`, I think it is an acceptable assumption. So in `Reflect.ownKeys` or any other methods we can't detect is this `Symbol` a primitive or an object. If they will be inherited from different prototypes, like in some other polyfills, for the correct work of many related cases, we should wrap `Object` constructor.

Issue #727, opened 2019-12-13
Maybe it is not a proper way to create object with `null` prototype with using Iframe


aleen42:
Recently, I have found that when I used IE11 to simulate IE8, and required `core-js/stable` in an iframe, it throwed an error specifiying that `iframe.contentWindow` was not asscessible when the script was run in an iframe, where scripts were not allowed to execute.

I set the iframe with `sandbox="allow-scripts"`, but it seemed that IE 11 did not acted like what I expected. In true IE8, `sandbox` is not worked, and there is no problem when accessing `iframe.contentWindow`. Therefore, is there any alternative to create `Object` with `null` prototype without using iframe?

zloirock:
Could you propose a better alternative? I don't see anything.

zloirock:
We could try to add a `ActiveXObject`-based workaround.

zloirock:
You could play with [it](https://github.com/es-shims/es5-shim/blob/ef78f316fee3eebc51a78f86f008bdac5b9790cf/es5-sham.js#L207-L242) and add a PR.

aleen42:
OK, I'll try it next weekday

PR #728, opened 2019-12-14
alternative for object.create under IE8 #727

aleen42:
PR for #727

### DIFF
--- a/packages/core-js/internals/object-create.js
+++ b/packages/core-js/internals/object-create.js
@@ -8,16 +8,27 @@ var sharedKey = require('../internals/shared-key');
 
 var IE_PROTO = sharedKey('IE_PROTO');
 var PROTOTYPE = 'prototype';
+var SCRIPT = 'script';
 var EmptyConstructor = function () { /* empty */ };
-
-// Create object with fake `null` prototype: use iframe Object with cleared prototype
-var NullProtoObject = function () {
-  // Thrash, waste and sodomy: IE GC bug
-  var iframe = documentCreateElement('iframe');
-  var length = enumBugKeys.length;
+var scriptTag = function (content) {
   var GT = '>';
   var LT = '<';
-  var SCRIPT = 'script';
+  return LT + SCRIPT + GT + content + LT + '/' + SCRIPT + GT;
+};
+
+// Create object with fake `null` prototype: use ActiveX Object with cleared prototype
+var NullProtoObjectViaActiveX = function (activeXDocument) {
+  activeXDocument.write(scriptTag(''));
+  activeXDocument.close();
+  var temp = activeXDocument.parentWindow.Object;
+  activeXDocument = null; // avoid memory leak
+  return temp;
+};
+
+// Create object with fake `null` prototype: use iframe Object with cleared prototype
+var NullProtoObjectViaIFrame = function () {
+  // Thrash, waste and sodomy: IE GC bug
+  var iframe = documentCreateElement('iframe');
   var JS = 'java' + SCRIPT + ':';
   var iframeDocument;
   iframe.style.display = 'none';
@@ -26,9 +37,24 @@ var NullProtoObject = function () {
   iframe.src = String(JS);
   iframeDocument = iframe.contentWindow.document;
   iframeDocument.open();
-  iframeDocument.write(LT + SCRIPT + GT + 'document.F=Object' + LT + '/' + SCRIPT + GT);
+  iframeDocument.write(scriptTag('document.F=Object'));
   iframeDocument.close();
-  NullProtoObject = iframeDocument.F;
+  return iframeDocument.F;
+};
+
+// Check for document.domain and active x support
+// No need to use active x approach when document.domain is not set
+// see https://github.com/es-shims/es5-shim/issues/150
+// variation of https://github.com/kitcambridge/es5-shim/commit/4f738ac066346
+// avoid IE GC bug
+var activeXDocument;
+var NullProtoObject = function () {
+  try {
+    /* global ActiveXObject */
+    activeXDocument = document.domain && new ActiveXObject('htmlfile');
+  } catch (error) { /* ignore */ }
+  NullProtoObject = activeXDocument ? NullProtoObjectViaActiveX(activeXDocument) : NullProtoObjectViaIFrame();
+  var length = enumBugKeys.length;
   while (length--) delete NullProtoObject[PROTOTYPE][enumBugKeys[length]];
   return NullProtoObject();
 };


packages/core-js/internals/object-create.js
```diff
@@ -45,6 +67,6 @@ module.exports = Object.create || function create(O, Properties) {
     EmptyConstructor[PROTOTYPE] = null;
     // add "__proto__" for Object.getPrototypeOf polyfill
     result[IE_PROTO] = O;
-  } else result = NullProtoObject();
+  } else result = viaActiveX() ? NullProtoObjectViaActiveX() : NullProtoObjectViaIFrame();
```
zloirock: It's the creation of 2 (!!!) ActiveX objects on each `Object.create(null)`. Also, `Object.prototype` from ActiveX is not cleared from properties.

aleen42: 👌 check the latest commit

packages/core-js/internals/object-create.js
```diff
@@ -10,11 +10,34 @@ var IE_PROTO = sharedKey('IE_PROTO');
...
+};
+
+// Create object with fake `null` prototype: use ActiveX Object with cleared prototype
+var NullProtoObjectViaActiveX = function () {
+  // Check for document.domain and active x support
+  // No need to use active x approach when document.domain is not set
+  // see https://github.com/es-shims/es5-shim/issues/150
+  // variation of https://github.com/kitcambridge/es5-shim/commit/4f738ac066346
+  try {
+    var activeXDocument = document.domain && new ActiveXObject('htmlfile');
```
zloirock: If it will throw an error, `ActiveXObject` will be created on each `Object.create(null)` even if required `IFrame`.

aleen42: And what is the suggestion? Create only one ActiveXObject for used?

aleen42: It seems there is the same question via IFrame

aleen42: check the latest commit

zloirock: Sure, it should be created only once.

packages/core-js/internals/object-create.js
```diff
@@ -10,11 +10,35 @@ var IE_PROTO = sharedKey('IE_PROTO');
...
+  } catch (e) {
+    return false;
+  }
+  activeXDocument.write('<script><\/script>');
```
zloirock: `<script><\/script>` can't be used directly even with the backslash since it will break some parsers and browsers when inlined to HTML. See IFrame.

aleen42: /wip

aleen42:
@zloirock Why can't I see the log of building failed above TravisCI?

zloirock:
I don't know. It's linting issues, you will have the same with local tests.

aleen42:
OK, I'll fix the problem of linting

packages/core-js/internals/object-create.js
```diff
@@ -26,13 +37,29 @@ var NullProtoObject = function () {
...
 };
 
+// Check for document.domain and active x support
+// No need to use active x approach when document.domain is not set
+// see https://github.com/es-shims/es5-shim/issues/150
+// variation of https://github.com/kitcambridge/es5-shim/commit/4f738ac066346
+var activeXDocument;
+try {
+  /* global ActiveXObject */
+  activeXDocument = document.domain && new ActiveXObject('htmlfile');
```
zloirock: Here should be used delayed initialization - we should not create `ActiveXObject` or `IFrame` if `Object.create` is available.

zloirock: If I understood correctly, it's a reason why currently tests fails - you try to create an `IFrame` on Node.

aleen42: Ooops...

aleen42: I just avoid initialization when `Object.create` is available

packages/core-js/internals/object-create.js
```diff
@@ -26,13 +37,27 @@ var NullProtoObject = function () {
...
+var NullProtoObject = Object.create /* avoid initialization */ || (function () {
+  try {
+    /* global ActiveXObject */
+    activeXDocument = document.domain && new ActiveXObject('htmlfile');
+  } catch (error) { /* ignore */ }
+  var ProtoObject = activeXDocument ? NullProtoObjectViaActiveX(activeXDocument) : NullProtoObjectViaIFrame();
+  var length = enumBugKeys.length;
+  while (length--) delete ProtoObject[PROTOTYPE][enumBugKeys[length]];
+  return ProtoObject;
+})();
```
zloirock: It would be better to use delayed initialization like it was before, like:
```js
var NullProtoObject = function () {
  // ...
  NullProtoObject = something;
  return NullProtoObject();
}
```

packages/core-js/internals/object-create.js
```diff
@@ -26,13 +37,27 @@ var NullProtoObject = function () {
...
-  while (length--) delete NullProtoObject[PROTOTYPE][enumBugKeys[length]];
-  return NullProtoObject();
+  return iframeDocument.F;
 };
 
+// Check for document.domain and active x support
+// No need to use active x approach when document.domain is not set
+// see https://github.com/es-shims/es5-shim/issues/150
+// variation of https://github.com/kitcambridge/es5-shim/commit/4f738ac066346
+var activeXDocument;
```
zloirock: Seems it's not required in this scope.

aleen42: When declared inside NullProtoObject, IE8 will throw an error, which message is around "Unable to run released script". So I have to declared globally

Use activeXDocument inside NullProtoObjectViaActiveX, rather than passing through parameters, should have the same problem.

zloirock:
Did you run tests on real IE8? Seems *sometimes* they are fails.

aleen42:
What detailed error have you caught? I have no real IE8 for karma to launch to test.

zloirock:
You could use VMs. You don't need karma for tests. Run `npm test` for usual CI and bundling all required files - and you could just open `tests/tests.html` file in IE8.

Details - when I tested, for some unknown reason in ~50% of the page loading some built-ins were unavailable. I don't know - it could be just the network problems or it's the problem with this script - since it's IE and specific DOM object, I hadn't enough time for exploring the problem. Little later I'll try to dig deeper.

aleen42:
image

I have the same error on the `master` branch

zloirock:
Interesting. I was able to reproduce it too but much less with 50% chance. However, I don't remember any changes that could affect IE8.

aleen42:
I can reproduce with 100% chance, I help you find out later.

zloirock:
@aleen42 how?

aleen42:
The same way, but caught `Reflect is not defined` any time

zloirock:
Already I can't reproduce it completely.
VirtualBox_8_18_12_2019_08_11_04

aleen42:
OK, I'll check it later.

aleen42:
What I need to test is `pure.html` rather than `tests.html`, as `pure.html` is ES3.

zloirock:
The same - no problems.

VirtualBox_8_18_12_2019_08_52_06

aleen42:
:ok_hand:

PR #732, opened 2019-12-17
Sticky RegExp 'y' and RegExp#test delegation

cvle:
This PR implements the sticky flag for Regular Expressions and `RegExp#test` delegation to `RegExp#exec`. It is based on the stale PR from @nicolo-ribaudo (many thanks!!) https://github.com/zloirock/core-js/pull/492.

Changes include:
- Rebased original work from @nicolo-ribaudo (Added `Co-authored-by` info in commit)
- Kept original implementation of `es.string.split`
- Implemented anchored sticky flag bahvior (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/sticky#Anchored_sticky_flag)
- Implemented `RegExp#test` delegation to `RegExp#exec`
- Adapted compat data and added tests

Tested in a multitude of browser versions using BrowserStack starting from the minimum supported versions of  `core-js`.

My goal was to make https://github.com/projectfluent/fluent.js work in IE11, which I've successfully verified! (also don't forget to use a different markup parser for those trying the same: https://gist.github.com/cvle/67b69517b7d3e83f5de9ccace270415c).

### DIFF
--- a/packages/core-js/es/regexp/sticky.js
+++ b/packages/core-js/es/regexp/sticky.js
@@ -1,0 +1,5 @@
+require('../../modules/es.regexp.sticky');
+
+module.exports = function (it) {
+  return it.sticky;
+};
--- a/packages/core-js/modules/es.regexp.constructor.js
+++ b/packages/core-js/modules/es.regexp.constructor.js
@@ -1,5 +1,4 @@
 var DESCRIPTORS = require('../internals/descriptors');
-var createNonEnumerableProperty = require('../internals/create-non-enumerable-property');
 var global = require('../internals/global');
 var isForced = require('../internals/is-forced');
 var inheritIfRequired = require('../internals/inherit-if-required');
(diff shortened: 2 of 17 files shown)


zloirock:
Since it works only for the global version, need to add empty files with similar names to `packages/core-js-pure/overwrite`.

packages/core-js/modules/es.regexp.constructor.js
```diff
@@ -33,13 +37,32 @@ if (FORCED) {
...
+      if (sticky) flags = flags.replace(/y/g, '');
+    }
+
+    var result = inheritIfRequired(
+      CORRECT_NEW ? new NativeRegExp(pattern, flags) : NativeRegExp(pattern, flags),
+      thisIsRegExp ? this : RegExpPrototype,
+      RegExpWrapper
+    );
+
+    if (UNSUPPORTED_Y) setInternalState(result, { sticky: sticky });
```
zloirock: Maybe it would be better to set only when `sticky` is `true`?

packages/core-js/es/regexp/sticky.js
```diff
@@ -0,0 +1,5 @@
+require('../../modules/es.regexp.sticky');
+
+module.exports = function (it) {
+  return it.sticky;
+};
```
zloirock: Also required `stable/regexp/sticky`, `features/regexp/sticky, the same for `RegExp#test` and adding them to `tests/commonjs`.

zloirock:
I think I'll fix it by myself. Thanks!

nicolo-ribaudo:
Thanks @cvle for working on my PR, and @zloirock for the quick fixes!

PR #744, opened 2020-01-08
fix corejs version type with babel-preset-env

jportella93:
Version should be a number https://babeljs.io/docs/en/babel-preset-env#corejs

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ import 'regenerator-runtime/runtime';
 
 #### `@babel/preset-env`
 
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
It should be fixed in the Babel docs. `"3.6"` is better than `3`, because `3` == `"3.0.0"`.

zloirock:
It could be a number, but much better to use a string and specify a minor version since with `corejs: 3` will not be injected modules added in minor `core-js@3` versions.

So `corejs: '3.6'` is the recommended way of configuring `core-js` in Babel. It would be better to improve docs on the Babel side.

jportella93:
Then Babel should change the config, rather than the docs, since it doesn't work with `corejs: '3.6'`. Just letting you guys know that there is a mismatch in the two repos docs which caused an error for me and wanted to avoid it for the next person who uses them.

zloirock:
> since it doesn't work with `corejs: '3.6'`

It works fine with `corejs: '3.6'` with installed `core-js@^3.6`.

PR #745, opened 2020-01-09
fixes #471

### LINKED ISSUE #471: Updates to String.prototype.replace doesn't handle $ in replacement value without capture group

mattclough1:
Fixes #471 in Safari. Not sure if this falls within the scope of core-js or not since the behavior of `$n` when n > m is supposed to be implementation-defined, which I think means Safari can do whatever they want?

### DIFF
--- a/packages/core-js/internals/fix-regexp-well-known-symbol-logic.js
+++ b/packages/core-js/internals/fix-regexp-well-known-symbol-logic.js
@@ -28,6 +28,14 @@ var REPLACE_KEEPS_$0 = (function () {
   return 'a'.replace(/./, '$0') === '$0';
 })();
 
+// Safari <= 13.0.3(?) substitutes nth capture where n>m with an empty string
+var REGEXP_REPLACE_SUBSTITUTES_UNDEFINED_CAPTURE = (function () {
+  if (RegExp.prototype['Symbol.replace']) {
+    return RegExp.prototype['Symbol.replace'].call(/./, 'a', '$0') === '';
+  }
+  return false;
+})();
+
 // Chrome 51 has a buggy "split" implementation when RegExp#exec !== nativeExec
 // Weex JS has frozen built-in prototypes, so use try / catch wrapper
 var SPLIT_WORKS_WITH_OVERWRITTEN_EXEC = !fails(function () {
@@ -75,7 +83,9 @@ module.exports = function (KEY, length, exec, sham) {
   if (
     !DELEGATES_TO_SYMBOL ||
     !DELEGATES_TO_EXEC ||
-    (KEY === 'replace' && !(REPLACE_SUPPORTS_NAMED_GROUPS && REPLACE_KEEPS_$0)) ||
+    (KEY === 'replace' && !(REPLACE_SUPPORTS_NAMED_GROUPS && (
+      REPLACE_KEEPS_$0 || REGEXP_REPLACE_SUBSTITUTES_UNDEFINED_CAPTURE
+    ))) ||
     (KEY === 'split' && !SPLIT_WORKS_WITH_OVERWRITTEN_EXEC)
   ) {
     var nativeRegExpMethod = /./[SYMBOL];
@@ -90,7 +100,10 @@ module.exports = function (KEY, length, exec, sham) {
         return { done: true, value: nativeMethod.call(str, regexp, arg2) };
       }
       return { done: false };
-    }, { REPLACE_KEEPS_$0: REPLACE_KEEPS_$0 });
+    }, {
+      REPLACE_KEEPS_$0: REPLACE_KEEPS_$0,
+      REGEXP_REPLACE_SUBSTITUTES_UNDEFINED_CAPTURE: REGEXP_REPLACE_SUBSTITUTES_UNDEFINED_CAPTURE
+    });
     var stringMethod = methods[0];
     var regexMethod = methods[1];
 
--- a/packages/core-js/internals/fix-regexp-well-known-symbol-logic.js
+++ b/packages/core-js/internals/fix-regexp-well-known-symbol-logic.js
@@ -30,8 +30,8 @@ var REPLACE_KEEPS_$0 = (function () {
 
 // Safari <= 13.0.3(?) substitutes nth capture where n>m with an empty string
 var REGEXP_REPLACE_SUBSTITUTES_UNDEFINED_CAPTURE = (function () {
-  if (RegExp.prototype['Symbol.replace']) {
-    return RegExp.prototype['Symbol.replace'].call(/./, 'a', '$0') === '';
+  if (RegExp.prototype[wellKnownSymbol('replace')]) {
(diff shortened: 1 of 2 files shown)


packages/core-js/internals/fix-regexp-well-known-symbol-logic.js
```diff
@@ -28,6 +28,14 @@ var REPLACE_KEEPS_$0 = (function () {
   return 'a'.replace(/./, '$0') === '$0';
 })();
 
+// Safari <= 13.0.3(?) substitutes nth capture where n>m with an empty string
+var REGEXP_REPLACE_SUBSTITUTES_UNDEFINED_CAPTURE = (function () {
+  if (RegExp.prototype['Symbol.replace']) {
```
zloirock: `Symbol.replace` is not a string. Use `wellKnownSymbol('replace')`.

zloirock:
I didn't explore the problem, I'll dig it a little later.

zloirock:
> Not sure if this falls within the scope of core-js or not since the behavior of $n when n > m is supposed to be implementation-defined, which I think means Safari can do whatever they want?

[In the spec I see](https://tc39.es/ecma262/#table-45):

> The nth element of captures, where n is a single digit in the range 1 to 9. If n ≤ m and the nth element of captures is undefined, use the empty String instead. **If n > m, no replacement is done.**

So yes, seems it should be fixed. Also, in Safari TP it works correctly.

packages/core-js/internals/fix-regexp-well-known-symbol-logic.js
```diff
@@ -28,6 +28,14 @@ var REPLACE_KEEPS_$0 = (function () {
   return 'a'.replace(/./, '$0') === '$0';
 })();
 
+// Safari <= 13.0.3(?) substitutes nth capture where n>m with an empty string
+var REGEXP_REPLACE_SUBSTITUTES_UNDEFINED_CAPTURE = (function () {
+  if (RegExp.prototype[wellKnownSymbol('replace')]) {
```
zloirock: I think it would be better to cache `wellKnownSymbol('replace')` as a constant, for example, `REPLACE`.

packages/core-js/internals/fix-regexp-well-known-symbol-logic.js
```diff
@@ -28,6 +28,14 @@ var REPLACE_KEEPS_$0 = (function () {
...
+    return RegExp.prototype[wellKnownSymbol('replace')].call(/./, 'a', '$0') === '';
```
zloirock: `/./[REPLACE]('a', '$0')`

packages/core-js/internals/fix-regexp-well-known-symbol-logic.js
```diff
@@ -75,7 +83,9 @@ module.exports = function (KEY, length, exec, sham) {
   if (
     !DELEGATES_TO_SYMBOL ||
     !DELEGATES_TO_EXEC ||
-    (KEY === 'replace' && !(REPLACE_SUPPORTS_NAMED_GROUPS && REPLACE_KEEPS_$0)) ||
+    (KEY === 'replace' && !(REPLACE_SUPPORTS_NAMED_GROUPS && (
+      REPLACE_KEEPS_$0 || REGEXP_REPLACE_SUBSTITUTES_UNDEFINED_CAPTURE
```
zloirock: A strange logic. Why `||`?

mattclough1:
> [In the spec I see](https://tc39.es/ecma262/#table-45):

Ah, shoot, yeah I was looking at an older spec 😕 Well I guess that just makes it a bug then!

packages/core-js/internals/fix-regexp-well-known-symbol-logic.js
mattclough1: My thinking was that `REPLACE_KEEPS_$0` is intended for lte IE11 but IE11 doesn't have the issue that `REGEXP_REPLACE_SUBSTITUTES_UNDEFINED_CAPTURE` intends to resolve and vice-versa, so the conditional would end up false for both of them?

packages/core-js/internals/fix-regexp-well-known-symbol-logic.js
zloirock: Also, I think that `=== ''` could be better to replace to `!== '$0'`, just in case.

packages/core-js/internals/fix-regexp-well-known-symbol-logic.js
zloirock: It's not something related to specific engines, it's related to specific bugs and implementations of old spec versions. At this moment we have 3 `replace`-specific criteria - so for me seems logically to use `&&` - `REPLACE_SUPPORTS_NAMED_GROUPS && REPLACE_KEEPS_$0 && !REGEXP_REPLACE_SUBSTITUTES_UNDEFINED_CAPTURE`.

zloirock:
At this moment, `fixRegExpWellKnownSymbolLogic` is the worst part of `core-js` - spaghetti, it seems required to refactor it in the near future...

packages/core-js/internals/fix-regexp-well-known-symbol-logic.js
nicolo-ribaudo: @mattclough1 Note that the `&&`s are wrapped inside `!(...)`, so the polyfill would be applied when _at least one_ of the tests is `false`

packages/core-js/internals/fix-regexp-well-known-symbol-logic.js
```diff
@@ -75,7 +84,11 @@ module.exports = function (KEY, length, exec, sham) {
   if (
     !DELEGATES_TO_SYMBOL ||
     !DELEGATES_TO_EXEC ||
-    (KEY === 'replace' && !(REPLACE_SUPPORTS_NAMED_GROUPS && REPLACE_KEEPS_$0)) ||
+    (KEY === 'replace' && !(
+      REPLACE_SUPPORTS_NAMED_GROUPS &&
+      REPLACE_KEEPS_$0 &&
+      REGEXP_REPLACE_SUBSTITUTES_UNDEFINED_CAPTURE
```
zloirock: Other criteria mean that the implementation is correct, but `REGEXP_REPLACE_SUBSTITUTES_UNDEFINED_CAPTURE` that it's buggy, so seem it should be `!REGEXP_REPLACE_SUBSTITUTES_UNDEFINED_CAPTURE`.

zloirock:
I'll finish it by myself, thanks for the PR.

Jan8787 (commented):
Oprava

Issue #756, opened 2020-01-23
IE/Edge missing in compat data


kroofy:
While browsing thru https://github.com/zloirock/core-js/blob/master/packages/core-js-compat/src/data.js

I noticed that `es.string.search` ([browser compat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/search#Browser_compatibility)) and `es.string.replace` among others are missing IE and Edge versions. Is this intentional or would it be welcomed by the project to have a PR submitted to add compat data for IE and Edge related polyfills? I would be happy to help it that's the case.

davidmerrique:
I also noticed that Edge is missing from a few other things, such as `es.promise`, `es.object.assign` (https://kangax.github.io/compat-table/es6/#test-Promise)

slowcheetah:
@kroofy by modern standards, `es.string.replace` & `es.string.search` should refer to well-known symbols. IE deviates from these standards and some others. Therefore, these methods should be polyfilled.

@davidmerrique see [#761](https://github.com/zloirock/core-js/pull/761#issuecomment-580740907)

PR #761, opened 2020-01-31
Include promise polyfill for IE 11

loginovma:
According to https://kangax.github.io/compat-table/es6/#test-Promise
we should include polyfill for promise in IE 11.

Mentioned in https://github.com/zloirock/core-js/issues/756


JakeChampion:
This file states the lowest version that supports the feature, which is no version of IE, which is why IE is not included in the objects for Promise polyfills.

loginovma:
Oh. ok, thanks! That was my wild guess why my promises won't get `.finally` in IE 11

Issue #755, opened 2020-01-21
Safari TP browser target breaks core-js-compat targets-parser


ehoogeveen-medweb:
I'm using require('core-js-builder') with the targets option set to a list of browsers. As of today "last 2 safari versions" expands to "safari 13" and "safari TP", which breaks core-js-compat with the following stack:

```
TypeError: Invalid Version: null
    at new SemVer (node_modules\core-js-compat\node_modules\semver\classes\semver.js:22:13)
    at compare (node_modules\core-js-compat\node_modules\semver\functions\compare.js:3:3)
    at lte (node_modules\core-js-compat\node_modules\semver\functions\lte.js:2:30)
    at cmp (node_modules\core-js-compat\node_modules\semver\functions\cmp.js:42:14)
    at compare (node_modules\core-js-compat\helpers.js:8:10)
    at module.exports (node_modules\core-js-compat\targets-parser.js:66:45)
    at module.exports (node_modules\core-js-compat\compat.js:28:25)
    at module.exports (node_modules\core-js-builder\index.js:35:26)
    at Object.<anonymous> (build.js:8:27)
    at Module._compile (internal/modules/cjs/loader.js:955:30)
```

I guess the non-numeric version number throws it off. I can work around the issue for now by changing "last 2 safari versions" to "safari 13".

nicolo-ribaudo:
It's a bug with `browserslist`: https://github.com/browserslist/browserslist/issues/439

ehoogeveen-medweb:
Ah I see, I did think it was strange for TP (tech preview?) to show up in that query. No need for a fix on my end then, though you can still trigger the error using something like "unreleased versions" or "unreleased safari versions" so I think it's a valid bug.

PR #785, opened 2020-03-21
Update dependencies

slowcheetah:


slowcheetah:
#755

scottarc:
The build failure should be detectable via the change in #786, if that helps

slowcheetah:
@scottarc thx. I will check it

Issue #797, opened 2020-04-06
Reduce output noise


delanni:
Hi!

I'm trying to build a website that helps poor young-adults in South-East Asia trying to find love during the coronavirus outbreak.
I'm running a free version of Jenknis, and my free version limit only allows a certain number of characters to be output on the console, before I get into a paywall.

I've recently added storybook to my toolkit to test my components with image comparison, and I started to see something very noisy in my console output, that takes away characters from my quota.
I'm seeing this as a part of my NPM install script:
```
npm WARN deprecated popper.js@1.16.1: You can find the new Popper v2 at @popperjs/core, this package is dedicated to the legacy v1

> core-js@3.6.4 postinstall /Users/frank/important/southeastasian-love-finder-during-virus-times/node_modules/@storybook/api/node_modules/core-js
> node -e "try{require('./postinstall')}catch(e){}"

Thank you for using core-js ( https://github.com/zloirock/core-js ) for polyfilling JavaScript standard library!

The project needs your help! Please consider supporting of core-js on Open Collective or Patreon: 
> https://opencollective.com/core-js 
> https://www.patreon.com/zloirock 

Also, the author of core-js ( https://github.com/zloirock ) is looking for a good job -)


> core-js@3.6.4 postinstall /Users/frank/important/southeastasian-love-finder-during-virus-times/node_modules/@storybook/channel-postmessage/node_modules/core-js
> node -e "try{require('./postinstall')}catch(e){}"


> core-js@3.6.4 postinstall /Users/frank/important/southeastasian-love-finder-during-virus-times/node_modules/@storybook/channels/node_modules/core-js
> node -e "try{require('./postinstall')}catch(e){}"


> core-js@3.6.4 postinstall /Users/frank/important/southeastasian-love-finder-during-virus-times/node_modules/@storybook/client-api/node_modules/core-js
> node -e "try{require('./postinstall')}catch(e){}"


> core-js@3.6.4 postinstall /Users/frank/important/southeastasian-love-finder-during-virus-times/node_modules/@storybook/client-logger/node_modules/core-js
> node -e "try{require('./postinstall')}catch(e){}"


> core-js@3.6.4 postinstall /Users/frank/important/southeastasian-love-finder-during-virus-times/node_modules/@storybook/components/node_modules/core-js
> node -e "try{require('./postinstall')}catch(e){}"


[...]

Issue #781, opened 2020-03-15
NPM install returning 1 after core-js Post-install in a Docker Environment  


dev-whiteflag:
Heyo,

I'm using Drone CI on a AWS Ubuntu to build a Angular project that depends on core-js; but don't know why core-js postinstall keeps stopping my building process, event with ``npm config set unsafe-perm true``.

Verbose isn't returning anything strange and my Docker image is node's latest, core-js's version is 3.6.4.

```
npm info lifecycle cookie-signature@1.0.6~postinstall: cookie-signature@1.0.6
npm info lifecycle copy-descriptor@0.1.1~postinstall: copy-descriptor@0.1.1
npm info lifecycle core-js@3.6.4~postinstall: core-js@3.6.4

> core-js@3.6.4 postinstall /drone/src/node_modules/core-js
> node -e "try{require('./postinstall')}catch(e){}"
```
Environment Variable CI is currently TRUE as well, there's anything more I've need to do?

dev-whiteflag:
Okay, I've tried running with `--silent` and it worked. Maybe is related to that CI var in the postinstall script. Don't know if Drone's env variables are in this filter.
https://github.com/zloirock/core-js/blob/4546ffeca298cbccf0843e31e8af6403693ab8ae/packages/core-js/postinstall.js#L13

ties-s:
I’m also having this issue. `--silent` doesn’t work for me.

slowcheetah:
Try this path #794

ties-s:
I’ve already tried with using the `CI` env var

Issue #767, opened 2020-02-12
State and governance of the project?


yvele:
----------

<br>

Edit: **The project still is alive**, some other contributors like @slowcheetah have permissions for the project to keep going, see https://github.com/zloirock/core-js/issues/767#issuecomment-603682034 👍 

Full summary of project governance here https://github.com/zloirock/core-js/issues/767#issuecomment-604892167 👍 

<br>

----------

Looks like @zloirock the author and [main maintainer](https://github.com/zloirock/core-js/graphs/contributors) of the project will be will be unavailable for ~~some time~~ 1.5 years.

Sources: https://github.com/zloirock/core-js/issues/767#issuecomment-598966371, https://github.com/zloirock/core-js/issues/757#issuecomment-579221001, https://github.com/zloirock/core-js/issues/747#issuecomment-573318269, https://github.com/zloirock/core-js/issues/548#issuecomment-494112872

What exactly is the state of the governance of this project?

The JavaScript community should be a bit concerned because @zloirock looks like to be the "only" maintainer. Does somebody else have admin privileges to write on this repo? Publish on npm and make this project not to die?

Or the only way is to "wait" for someone to fork this repo? ~~Maybe someone from @babel~~ (poking @nicolo-ribaudo and @danez 🤷‍♂). Looks like @babel [doesn't have bandwith to fork this project](https://github.com/zloirock/core-js/issues/767#issuecomment-600839713).

A huge open source project ([25M weekly downloads](https://www.npmjs.com/package/core-js)) like this should be maintained by more than a single person 🤔 

Any clues on [the](https://github.com/babel/babel/issues/10008) [future](https://github.com/zloirock/core-js/issues/496) of this project?

PS: I don't know your personal story @zloirock but I'm grateful for your amazing work on this project.. hoping everything will be fine

Edit: ~~This project is dead, see https://github.com/zloirock/core-js/issues/767#issuecomment-598966371~~

ashpr:
@zloirock  Making himself the only maintainer was extremely poor handling of such a well used repo.. but I can't say I'm surprised. He's been extremely protective of it.

I think, in time, this project may need to be forked.

delanni:
Looks like the Karma Police got him...

danielrree:
[...]

Issue #757, opened 2020-01-24
Please remove the post install script


pronebird:
Hi,

It's absolutely terrific that you're looking for a job, but please do not do this via `postinstall` script. There are job boards, meetups where you can connect with others, LinkedIn, and so forth. 

```
>> node -e "try{require('./postinstall')}catch(e){}"
```

Cheers,
Andrej

veeral-patel:
@pronebird Hey! Looking at other issues in the repository, the maintainer has said that he won't remove the postinstall message, to spite people who want to remove it and maybe for financial reasons as well -- he's mentioned he'll be going to prison in a couple days.

Correct me if I'm wrong @zloirock

ljmerza:
where did you see this? If that's true I hope someone forks this projects and takes it from him

veeral-patel:
https://github.com/zloirock/core-js/issues/747#issuecomment-573318269

https://github.com/zloirock/core-js/issues/548#issuecomment-494112872

@ljmerza

veeral-patel:
@pronebird agree with your original comment. I don't mind the postinstall message, considering all the work he's put into this library for free, but I worry that continuously saying he wants a good job undermines him. I'd expect he be in high demand considering he's authoring a standard library for Javascript, no small undertaking, and I hope this doesn't hurt him.

ljmerza:
well considering he was committing on a almost daily basis then went silent 12 days ago i would imagine this project just died. here's to hoping babel or someone else takes this over. he sounded like a real winner

Kein:
> [#747 (comment)](https://github.com/zloirock/core-js/issues/747#issuecomment-573318269)
> [#548 (comment)](https://github.com/zloirock/core-js/issues/548#issuecomment-494112872)
> @ljmerza

Since he is from Russia - I dont think he was joking at all.

slowcheetah:
@Kein this is not a joke.
@zloirock will be unavailable for some time. 
[Here](https://github.com/zloirock/core-js/issues/548#issuecomment-494112872) he clearly expressed his opinion on this issue.
Closed.

Daniel15:
> well considering he was committing on a almost daily basis then went silent 12 days ago i would imagine this project just died. here's to hoping babel or someone else takes this over. he sounded like a real winner

[...]

PR #806, opened 2020-04-12
limit postinstall ad to show only one time

obnoxiousnerd:
When installing multiple packages that depend on core-js, the postinstall message will pop up mutiple times which will add unnecessary verbosity to the terminal log. I know #767, #548, #797, #757, #781 ,#729, #708, refer to this problem and since @zloirock refuses to remove this, lets limit it to one time so that verbosity gets truncated to the least.

### DIFF
--- a/packages/core-js/postinstall.js
+++ b/packages/core-js/postinstall.js
@@ -4,7 +4,7 @@ var os = require('os');
 var path = require('path');
 var env = process.env;
 
-var AD_SHOWN = is(env.AD_SHOWN)
+var AD_SHOWN = is(env.AD_SHOWN);
 var ADBLOCK = is(env.ADBLOCK);
 var COLOR = is(env.npm_config_color);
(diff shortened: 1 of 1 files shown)


sgammon:
@obnoxiousnerd great work

obnoxiousnerd:
My pleasure, @sgammon .
Since the author of this project, @zloirock can't have work on this repo and since @slowcheetah only has write access to the repository, so I kindly request them to add some more collaborators so that the project survives.

slowcheetah:
@obnoxiousnerd I see this PR. I am ready to accept it. 
Today I will have the opportunity to communicate with @zloirock . After that I will merge PR.

obnoxiousnerd:
All right, @slowcheetah

forrestblade:
Just doooo it

slowcheetah:
Repeat ad protection already exists. If the problem arises, then most likely the old version of core-js is installed.

PR #841, opened 2020-06-19
fix: some runtime can't use new Function

DiamondYuan:
There are some customized environments. New Function are not allowed .

this code is from regenerator

https://github.com/facebook/regenerator/blob/3db5564ff18e3f5bdafc4577522b41a006cc2c91/packages/regenerator-runtime/runtime-module.js#L10


zloirock:
Thanks, but it will not work in many cases since too many tools add the strict mode, and instead of `globalThis` this code will return `undefined`. There were many discussions about it. [A link to one of them](https://github.com/zloirock/core-js/issues/86#issuecomment-115759028) you could find in the file which you wanted to fix.

PR #842, opened 2020-06-28
move `String#replaceAll` to stable ES

pustovalov:
per Jun TC39 meeting

https://github.com/tc39/proposal-string-replaceall


pustovalov:
Do I need to add `modules/es.symbol.replace-all` to `packages/core-js/es/symbol/index.js`?

zloirock:
Thanks.

PR #859, opened 2020-08-28
Add Android 81

Hypnosphi:
```
 % npx browserslist 'last 1 android version'
android 81
```

### DIFF
--- a/packages/core-js-compat/src/mapping.js
+++ b/packages/core-js-compat/src/mapping.js
@@ -91,6 +91,7 @@ module.exports = {
     [12, '4.0'],
     [30, '4.4'],
     [33, '4.4.3'],
+    [81, '81'],
   ],
   ChromeToElectron,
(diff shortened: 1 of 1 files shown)


zloirock:
Thanks for the PR, but in `core-js` `android > 4.4.4` [already handled as the related chrome version](https://github.com/zloirock/core-js/blob/ebddf97e642c37257e9ef8f6359ecea3226da253/packages/core-js-compat/targets-parser.js#L54-L56).

Hypnosphi:
@zloirock babel packages [use](https://github.com/babel/babel/search?q=core-js-compat)  `core-js-compat/data` directly. What do you think is the best course of actions for them in order to support modern android?

zloirock:
I think that it should be resolved on the Babel targets parser side. Makes no sense to have a separate field in compat data for modern Android since the modern Android browser is just Chrome.

zloirock:
Ok, I'll move it directly to compat data for consistency and preventing questions why old android web view in compat data json, but modern - not.

PR #863, opened 2020-09-18
fix: resolve postMessage take a error on ios when protocol is file

hlerenow:
On my ios device, when I run html with file protocol, the console will get an error,
image

Although [#771 ] has resolve this problem,  I think there has some other issue, because ```!fails(post)``` will run before ```location.protocol !== 'file:'``` ，and ```post``` method will be call,  then postMessage will be call. There was a problem
```
var post = function (id) {
  // old engines have not location.origin
  global.postMessage(id + '', location.protocol + '//' + location.host);
};
```

so I switch this position

### DIFF
--- a/packages/core-js/internals/task.js
+++ b/packages/core-js/internals/task.js
@@ -80,8 +80,8 @@ if (!set || !clear) {
     global.addEventListener &&
     typeof postMessage == 'function' &&
     !global.importScripts &&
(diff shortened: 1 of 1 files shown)


zloirock:
`fails` call this code in `try / catch`, so it should work without any problems. However, thanks, I'll accept your PR since this order looks more logical.

PR #875, opened 2020-10-27
CI: Travis CI -> Github actions

pustovalov:
Github Actions
Done in 3m 37s
https://github.com/pustovalov/core-js/actions/runs/330138788

Travis CI
Ran for 2 hrs 3 min 30 sec
Total time 29 min 38 sec
https://travis-ci.org/github/zloirock/core-js/builds/739077045

image


zloirock:
Thanks. I thought about it, but I'm still not completely decided how it should look. Let's delay it for some time.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,34 @@
...
+on: [push, pull_request]
+
+jobs:
+  build:
+    name: Specs
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        node: [10.x, 12.x, 14.x]
```
zloirock, line 12: Instead of `.x` here should be first LTS versions. And here should be commented `15` which should be uncommented after fixing an NPM bug or adding a fallback on the `core-js` side.

pustovalov: Example:
https://github.com/pustovalov/core-js/actions/runs/332170920
https://github.com/pustovalov/core-js/actions/runs/332170917

zloirock:
Now is almost impossible to use Travis, so merged. Thanks -)

PR #881, opened 2020-11-03
[compat] IE10 supports String.prototype.trim()

apepper:
`String.prototype.trim()` is supported by IE10 or newer.

trim on IE10

See https://github.com/mdn/browser-compat-data/pull/7254 for MDN data related to this.

### DIFF
--- a/packages/core-js-compat/src/data.js
+++ b/packages/core-js-compat/src/data.js
@@ -929,6 +929,7 @@ const data = {
     chrome: '59',
     edge: '15',
     firefox: '52',
+    ie: '10',
     safari: '12.1',
   },
   'es.string.trim-end': {


JakeChampion:
@apepper IE10 only supports the old `String.prototype.trim` definition, where `"\u180E".trim().length` would return `0` instead of returning `1`.

>In ECMASCript 2016, Unicode 8.0.0 or higher is mandated, as opposed to ECMAScript 2015 which mandated Unicode 5.1. In particular, this caused U+180E MONGOLIAN VOWEL SEPARATOR, which was in the Space_Separator (Zs) category and thus treated as whitespace in ECMAScript 2015, to be moved to the Format (Cf) category (as of Unicode 6.3.0). This causes whitespace-sensitive methods to behave differently. For example, "\u180E".trim().length was 0 in previous editions, but 1 in ECMAScript 2016 and later. Additionally, ECMAScript 2017 mandated always using the latest version of the Unicode standard.
>
> [ECMA262 - F Additions and Changes That Introduce Incompatibilities with Prior Editions](https://tc39.es/ecma262/#sec-additions-and-changes-that-introduce-incompatibilities-with-prior-editions)

zloirock:
Thanks, but @JakeChampion is correct.

`core-js` compat data based on `core-js` feature detection. [Here](http://raw.githack.com/zloirock/core-js/master/tests/compat/index.html) you could check which modules required in the target browser.

PR #882, opened 2020-11-08
refactor(core-js-builder): change blacklist term

rivajunior:
blacklist is racial offensive and can easely be changed by other terms as `exclude` or `denylist`.

I've choose `exclude` because I think it fits better than denylist in this context. But I can change it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,7 +321,7 @@ It does not work with some features. Also, if you change the default behaviour, 
 
 ### Custom build
 
-For some cases could be useful adding a blacklist of features or generation a polyfill for target engines. You could use [`core-js-builder`](/packages/core-js-builder) package for that.
+For some cases could be useful adding a _exclude_ of features or generation a polyfill for target engines. You could use [`core-js-builder`](/packages/core-js-builder) package for that.
 
 ### Compatibility data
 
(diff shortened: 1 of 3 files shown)


zloirock:
I don't think that `blacklist` is racial offensive - I can't imagine what kind of complexes a person should have in order to be offended by this. However, yes, for this case `exclude` sounds better and I could accept this PR. But it's a breaking change, so required a fallback to `blacklist`.

README.md
```diff
@@ -114,29 +118,34 @@ Promise.resolve(32).then(x => console.log(x)); // => 32
...
+### Installation
+
+```sh
+# global version
 npm install --save core-js@3.7.0
-// version without global namespace pollution
+# version without global namespace pollution
 npm install --save core-js-pure@3.7.0
-// bundled global version
+# bundled global version
```
zloirock, line 129: Why it's changed? Please, remove all changes unrelated to `exclude` from this PR.

README.md
```diff
@@ -321,7 +321,7 @@ It does not work with some features. Also, if you change the default behaviour,
 
 ### Custom build
 
-For some cases could be useful adding a blacklist of features or generation a polyfill for target engines. You could use [`core-js-builder`](/packages/core-js-builder) package for that.
+For some cases could be useful adding a _exclude_ of features or generation a polyfill for target engines. You could use [`core-js-builder`](/packages/core-js-builder) package for that.
```
zloirock, line 324: Please, rephrase into something more readable. The rest entries - too.

rivajunior: @zloirock what do you think about the following:


> For some cases could be useful generating a especific polyfill for target engines.
>
> You can use [`core-js-builder`](/packages/core-js-builder) package for include or exclude parts of `core-js`. And optionally generate a file from that.

PR #889, opened 2020-11-15
Add relative indexing method proposal

zloirock:
https://github.com/tc39/proposal-relative-indexing-method


zloirock:
@syg @tabatkins hey, how do you think, makes it sense to add this proposal to `core-js` with the current name / in the current form? The conflict with [another proposal](https://github.com/mathiasbynens/String.prototype.at), possible problems and solutions see [here](https://github.com/tc39/proposal-relative-indexing-method/issues/33#issuecomment-728954227).

mathiasbynens:
Yes, the old proposal should not be considered at all IMHO.

zloirock:
@mathiasbynens sure, in this case, `core-js` will use `String#at` from the new proposal in the next major version. But the problem here is that for many years was available the proposal with code points semantic, most likely someone uses one of our polyfills in production (billions of `core-js` copies with this polyfill for 6+ years - it's too long, and not all of the users understand what means stage 0 proposals), and after adding the method with the same name and code units semantic it could break the web. So, I think, here could help or usage code points semantic, or changing the name. Or just ignore it, save it as is, and wait for the feedback from the early browser implementations?

syg:
Hm, this is indeed a risk with shipping early-stage proposals. I also think the old proposal should be removed in favor of the new one. A method named `at` with code point semantics, however, is explicitly rejected as part of the current relative indexing proposal, to be consistent with other indexables.

To make sure I understand the risk, `core-js`'s old polyfill for the old proposal only adds `String#at` if one doesn't already exist, so if one suddenly existed (because engines shipped it), then that might break? If the browsers get such bug reports, then the next name to try is `itemAt`, but I'd like to wait for the bug reports first.

zloirock:
> To make sure I understand the risk, `core-js`'s old polyfill for the old proposal only adds `String#at` if one doesn't already exist, so if one suddenly existed (because engines shipped it), then that might break?

Yep, without any additional code units / code points feature detection since I didn't think that in the future the method with this name will be added with code units semantic. It will be added in the new patch releases, however, there are many projects where the old `core-js` version will not be updated.

PR #892, opened 2020-12-01
Change variable name from domain to domainRef

gaastonsr:
We are upgrading an application to Node 12 and we started getting the following error: `TypeError: domain.enter is not a function`.

Apparently, Node.js has issues handling variable names with the name `domain`. It sometimes confuses the local variable with the legacy domain object.

Here's some prior art about that:

- https://github.com/phillipj/gitlab-search/issues/9#issuecomment-549074757
- https://github.com/nodejs/node/issues/34720


gaastonsr:
We found the root cause. We were polluting `req.domain` in our express server. This won't be needed anymore.

PR #911, opened 2021-02-14
use a smaller value for "length" when testing array "ToLength"

terencehonles:
Apologies if I don't understand why -1 was chosen in the first place, but I believe this change is appropriate. This bug is triggered when https://github.com/wkhtmltopdf/wkhtmltopdf (Qt Webkit) tries to load `packages/core-js/modules/es.array.splice.js` (it may also be triggered on `index-of.js`). It looks like how it handles `Object.defineProperty` does not interrupt execution and for splice it is trying to delete a ton of properties `[2..-1]` before eventually throwing (I believe it does actually throw, but it has to do a lot of work before getting to that point, shortening the array allows it to throw sooner).

Commit description:

Using the value -1 for the length of the object in the `arrayMethodUsesToLength` tests causes implementations to test against a very large "array". Use a smaller value to not cause those implementations to hang. This change allows all the uses of `arrayMethodUsesToLength` to use the actual function they are testing rather than a "proxy". In addition, change the testing function to use `apply` rather than `call` so that implementations that count the number of arguments will use the default values for any arguments that are optional rather than trying to interpret `undefined` incorrectly.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,6 +7,7 @@ jobs:
     name: tests
 
     strategy:
+      fail-fast: false
       matrix:
         os: [
           'ubuntu-latest',
--- a/packages/core-js/internals/array-last-index-of.js
+++ b/packages/core-js/internals/array-last-index-of.js
@@ -9,8 +9,7 @@ var min = Math.min;
 var nativeLastIndexOf = [].lastIndexOf;
 var NEGATIVE_ZERO = !!nativeLastIndexOf && 1 / [1].lastIndexOf(1, -0) < 0;
 var STRICT_METHOD = arrayMethodIsStrict('lastIndexOf');
-// For preventing possible almost infinite loop in non-standard implementations, test the forward version of the method
-var USES_TO_LENGTH = arrayMethodUsesToLength('indexOf', { ACCESSORS: true, 1: 0 });
+var USES_TO_LENGTH = arrayMethodUsesToLength('lastIndexOf', { ACCESSORS: true });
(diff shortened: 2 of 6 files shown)


terencehonles:
I can't tell if windows is flaky or just broken on 10.13 and 14.15 (12.13 passed), but the breakage should not be related to this change and all the other versions of the tests look like they are passing (a couple left to finish)

zloirock:
CI tests on windows are broken because of an NPM issue and are not related to your PR.

zloirock:
You could rebase it, I added a workaround for this NPM bug.

However, I don't think that it's a correct fix for this problem.

By the current spec, `Array` methods should apply [`ToLength`](https://tc39.es/ecma262/#sec-tolength) to `.length` of arrays, and `length: -1` should be converted to `0`. But some old engines use `ToUint32` or other conversions and `-1` converted to a huge integer. We should somehow detect it - and we use `defineProperty` with `throw` in `array-method-uses-to-length` for that. But seems, in this case, that intersects with `defineProperty` bug in old WebKit or something like that.

zloirock:
I dropped `ToLength` detection from array methods feature detection since it could cause hanging of some old engines and I don't see any possible workaround which could work everywhere.

terencehonles:
Sounds good, thanks for the change.

terencehonles:
@zloirock QQ, should the change be cherry picked to the `v3` branch? Not sure the difference between `v3` and `master` or will that happen later?

zloirock:
@terencehonles nope, it's not required. `v3` vas a development branch before `core-js@3` release and will be a legacy branch after `core-js@4` release.

PR #913, opened 2021-02-22
[FIX]: fix text error

roadupcc:
fix text error


zloirock:
I don't speak Chinese and I don't see any serious difference using automatic translation. Could you explain those changes?

roadupcc:
@zloirock 

"幼稚的" means " lack of thinking about something" , so this "naive " should be understood as "简单的" ( simple )
"喝" means "drink" , "和" means "and" , so this is an error


> I don't speak Chinese and I don't see any serious difference using automatic translation. Could you explain those changes?

zloirock:
Ok, thanks.

Issue #810, opened 2020-04-13
core-js@3.6 breaks xregexp patterns on IE11


psmyrdek:
Related to: https://github.com/angular/angular-cli/issues/17375

There are two issues with XRegExp and core-js I noticed recently:

**1)** On `core-js@3.6.4` this script causes endless loop:
```js
import 'core-js/features/regexp';
import XRegExp from 'xregexp';
window['XRegExp'] = XRegExp; // only to test some patterns in dev tools console
```
When using `core-js@3.5.0`  everything works fine.

**2)** Pattern that is working fine in `core-js@3.5.0` now breaks (the `\s`, whitechar range, is gone)

I'm trying to test the following code:
```js
import XRegExp from 'xregexp';
window['XRegExp'] = XRegExp;

console.log(XRegExp('(^((?=[uuuuuuuuu\\sxxxxxxxxx\\p{L}\\p{N}\\p{P}\\p{S}\\p{M}/])[^;\\\\<>{}])*$)|(^$)'))
console.log(XRegExp('(^((?=[uuuuuuuuu\\sxxxxxxxxx\\p{L}\\p{N}\\p{P}\\p{S}\\p{M}/])[^;\\\\<>{}])*$)|(^$)').test('test me'));
```
And I see two different regexp patterns that got built by xregexp:

![image](https://user-images.githubusercontent.com/6187417/79115278-ba6c3f00-7d85-11ea-8880-0afd4dfd7cc8.png)

![image](https://user-images.githubusercontent.com/6187417/79115272-b93b1200-7d85-11ea-855d-10e5af872f74.png)

kassenov:
We experienced the described issue on IE11 with version 3.6, when reverted to 3.5.0 it works.

psmyrdek:
One solution may be to use https://www.npmjs.com/package/babel-plugin-transform-xregexp and build patterns in build time. Haven't tested it properly yet - feedback welcomed.

walkerdb:
Seeing the same issue as well

steveworkman:
Looks like a duplicate of #751

kabbany:
Still having the same issues. Any updates regarding the issue ?

zloirock:
Seems on `xregexp` side [added a workaround](https://github.com/slevithan/xregexp/commit/bece2eb1b4e4206ac07c62d75c457effd93eb491), but it should be fixed on `core-js` side.

zloirock:
#754 related

[...]

PR #919, opened 2021-03-19
fix multiple usages of regexp-exec

### LINKED ISSUE #751: IE11 + core-js@3.6.4 String.prototype.split doesn't work with regexp
Screen Shot 2020-01-15 at 9 40 26 PM

repo to reproduce
https://github.com/kmsheng/core-js-bug

Note that core-js was imported twice,
one in src/js/index.js and the other one in assets/js/default.js

simlar issue
https://github.com/zloirock/core-js/issues/741

Khartir:
Closes #751

I would have liked to add a test for this, but I have honestly no idea of how to do so.

### DIFF
--- a/packages/core-js/internals/regexp-exec.js
+++ b/packages/core-js/internals/regexp-exec.js
@@ -1,6 +1,9 @@
 'use strict';
 var regexpFlags = require('./regexp-flags');
 var stickyHelpers = require('./regexp-sticky-helpers');
+// Ensure sticky flag is correctly patched
+require('../modules/es.regexp.constructor');
+require('../modules/es.regexp.sticky');
 
 var nativeExec = RegExp.prototype.exec;
 // This always refers to the native implementation, because the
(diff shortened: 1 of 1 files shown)


zloirock:
I need some time to explore this issue deeper.

zloirock:
`core-js` should be maximally modular and `es.regexp.exec` module should work without `es.regexp.constructor` or `es.regexp.sticky`.

zloirock:
I'm confused in this code. @nicolo-ribaudo could you say something about #751 and #810?

Khartir:
@zloirock As far as I can tell, the problem with making `es.regexp.exec`, `es.regexp.constructor` and `es.regexp.sticky`  independent of each other is that in order to support the `y` flag all have to be modified: the constructor to accept the flag, sticky to add the property and exec to actually handle it.

So maybe the solution should be to combine the modules into one? The only way to split this cleanly in different packages would probably be to have a `sticky` package that would modify constructor, property and exec all on its own and have the other packages only do what they did before. That would require overwriting the constructor and exec multiple times (in the correct order) and would still not solve #751 because [this comparison](https://github.com/zloirock/core-js/blob/f10bf27fa665cde29e549262da977946557c84b4/packages/core-js/internals/fix-regexp-well-known-symbol-logic.js#L81) would still fail.

zloirock:
@Khartir too many other modules depend on `es.regexp.exec`, we should not enforce load `es.regexp.constructor` and `es.regexp.sticky` to modules that depend on it. So we should fix it without uniting those modules.

zloirock:
I'll replace the check of `regexpExec` to `RegExp.prototype.exec` in `maybeCallNative` as a temporal fix - it will fix this issue, but `RegExp` -> `String` logic should be completely reworked.

PR #928, opened 2021-04-21
Update hasOwn to reflect change in operation order

jamiebuilds:
`ToObject` is now called before `ToPropertyKey` https://github.com/tc39/proposal-accessible-object-hasownproperty/commit/d74b1776d54e7e7ca070bba749b63cebb51d2aaf

I'm not sure if this is the right place to do this since its an internal method

### DIFF
--- a/packages/core-js/internals/has.js
+++ b/packages/core-js/internals/has.js
@@ -1,5 +1,7 @@
+var toObject = require('../internals/to-object');
+
 var hasOwnProperty = {}.hasOwnProperty;
 
 module.exports = function hasOwn(it, key) {
-  return hasOwnProperty.call(Object(it), key);
+  return hasOwnProperty.call(toObject(it), key);
 };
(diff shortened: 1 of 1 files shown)


zloirock:
See https://github.com/tc39/proposal-accessible-object-hasownproperty/pull/7#discussion_r617781246

`Object` is not an equal of `ToObject`.

Since `has` helper is used in some other place, I think that it's better to update `esnext.object.has-own` to something like:
```js
var $ = require('../internals/export');
var hasOwn = require('../internals/has');
var toObject = require('../internals/to-object');

// `Object.hasOwn` method
// https://github.com/tc39/proposal-accessible-object-hasownproperty
$({ target: 'Object', stat: true }, {
  hasOwn: function hasOwn(object, property) {
    return hasOwn(toObject(object), property);
  }
});
```

zloirock:
I thought a little more and I think that it's fine just add `RequireObjectCoercible` check to `internals/has` helper.
```js
var toObject = require('../internals/to-object');

var hasOwnProperty = {}.hasOwnProperty;

module.exports = function hasOwn(it, key) {
  return hasOwnProperty.call(toObject(it), key);
};
```

PR #942, opened 2021-06-10
Add compat data for rhino JS engine

gausie:
To get this running on the current version of Rhino I needed to:

* Change `GLOBAL` in `tests.js` to equal `global`
* Replace `console.log` with `print`
* Return `false` for "es.string.search" because the code causes an NPE

These (in particular number 3) are good little tasks to fix upstream in Rhino though

Then it ran no problem with `java -jar rhino.jar -version 200 -require ./node-runner.js`

### DIFF
--- a/packages/core-js-compat/src/data.js
+++ b/packages/core-js-compat/src/data.js
@@ -21,23 +21,27 @@ const data = {
     edge: '15',
     firefox: '49',
     safari: '10.0',
+    rhino: '1.7',
   },
   'es.symbol.is-concat-spreadable': {
     chrome: '48',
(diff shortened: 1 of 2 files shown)


zloirock:
Thanks -)

zloirock:
The question: what means `flag` here http://mozilla.github.io/rhino/compat/engines.html?

gausie:
It means running it with the es6 flag - its misleading because its not a pre-release flag, its a flag that the docs recommend everyone set when working with Rhino.

zloirock (approved):

PR #943, opened 2021-06-11
Specify patch version for rhino

gausie:
Rhino hasn't bumped a minor version in over a decade and all the changes happen on `1.7.x`. Accordingly, I think it might be best to include that in the data, particularly because I expect a new version under 1.7 will come out soon with higher ES6 compatibility.


zloirock:
Thanks, but I made it just a minute ago https://github.com/zloirock/core-js/commit/5a98661ca7254bbe7aa3e9bbc55cf89adb823d86 -)

PR #959, opened 2021-07-15
Add array grouping proposal

zloirock:
https://github.com/tc39/proposal-array-grouping


zloirock:
@jridgewell could you review it?

packages/core-js/internals/array-group-by.js
```diff
@@ -0,0 +1,33 @@
...
+  var target = objectCreate(null);
+  var length = toLength(self.length);
+  var index = 0;
+  var Constructor, key, value;
+  for (;length > index; index++) {
+    value = self[index];
+    key = toPropertyKey(boundFunction(value, index, O));
+    // in some IE10 builds, `hasOwnProperty` returns incorrect result on integer keys
+    // but since it's a `null` prototype object, we can safely use `in`
+    if (key in target) push.call(target[key], value);
```
jridgewell, line 24: Nit: I switched a List type in the current spec, but I don't know how you implement list (it may just be a normal array?) But I think this exposes set operations in an incorrect order?

jridgewell (approved):

packages/core-js/internals/array-group-by.js
zloirock: Lists are unobservable, so it's optimized for direct adding plain arrays to the `null` prototype object. If it's a custom array constructor or a typed array, they are created on a second loop pass and all observable operations evaluated in the correct order. The only observable difference with the current spec draft is [an optimization of getting `@@species` constructor](https://github.com/tc39/proposal-array-grouping/issues/11).

zloirock: Sure, it's possible to define setters on `Array.prototype` and observe it, but I think that it's an acceptable limitation. This is a limitation of all polyfills of features that use `List` that I know. It's possible to use arrays with `null` prototype or objects with `null` prototype, but I don't think that it's worth it.

PR #966, opened 2021-07-29
Fix crash in `NullProtoObjectViaIFrame()` on WSH

gnh1201:
If you modify this part, it works well in WSH.

If `NullProtoObject()` fails, create an `htmlfile` object immediately without checking `document.domain`.


zloirock:
Thanks, I'll adapt it by myself.

PR #967, opened 2021-08-02
Fix microtasks on iOS Pebble

leo60228:
The iOS Pebble app bundles its own copy of JavaScriptCore. This version includes a Promise constructor, but it's completely broken. The Promise polyfill detects this, but microtasks try to use it anyway.


zloirock:
Ok, thanks.

PR #981, opened 2021-09-05
BigInt support

Rudxain:
arr[n] where `arr` is an Array and `n` is a BigInt, is supported by Chromium. arr.at(n) throws an error because `toInteger` doesn't support BigInts

### DIFF
--- a/packages/core-js/modules/es.array.at.js
+++ b/packages/core-js/modules/es.array.at.js
@@ -11,8 +11,8 @@ $({ target: 'Array', proto: true }, {
   at: function at(index) {
     var O = toObject(this);
     var len = toLength(O.length);
-    var relativeIndex = toInteger(index);
-    var k = relativeIndex >= 0 ? relativeIndex : len + relativeIndex;
+    var relativeIndex = toInteger(Number(index)); //Number() supports BigInt
+    var k = relativeIndex + (relativeIndex < 0 && len);
     return (k < 0 || k >= len) ? undefined : O[k];
   }
 });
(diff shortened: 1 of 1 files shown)


zloirock:
> arr[n] where arr is an Array and n is a BigInt, is supported by Chromium

Chromium 95:
image
Chrome 93:
image
FF Nightly:
image
FF91:
image
Spec draft:
https://tc39.es/proposal-relative-indexing-method/
Spec PR:
https://github.com/tc39/ecma262/pull/2493/files

Everywhere it throws an error on bigints. Why do you think that it should support BigInt?

Rudxain:
If the `at` method didn't support BigInts, then then one of the goals of the proposal would be incomplete (I'm talking about conciseness). Developers would have to write `array.at(Number(x))` just to have similar behavior to `array[array.length - Number(x)]` . The following is a modification I made of the polyfill, it works on MS Edge Chromium (I haven't tested EdgeHTML) Version 93.0.961.47:
`//github.com/tc39/proposal-relative-indexing-method#polyfill
function at(n) {
   n = Math.trunc(Number(n)) || 0; if (n < 0) n += this.length;
   return n < 0 || n >= this.length ? undefined : this[n]
};
const TypedArray = Reflect.getPrototypeOf(Int8Array);
for (const C of [Array, String, TypedArray]) {
   Object.defineProperty(C.prototype, "at", {value: at, writable: true, enumerable: false, configurable: true})
};`


And I apologize for replying late, I didn't read my mail and didn't get a notification.

zloirock:
`core-js` is a polyfill and should follow the spec. It's an issue for the proposal, you could write it here https://github.com/tc39/proposal-relative-indexing-method

Rudxain:
Thanks for the info and your time. I didn't know I had to post it there. I'm relatively new to GitHub, sorry

PR #987, opened 2021-09-21
remove `forEach` from majority of DOM collection prototypes

moorejs:
Issue #988.

An alternative to this PR is to expand the compatibility test to include all the DOM Iterables.

I wasn't sure how you would want this change, so I have an simpler implementation [here](https://github.com/zloirock/core-js/commit/bd145f066acd096340a1361867935719d6e753e8).


chenyulun:
Why not remove it from 4.x.x？

zloirock:
The comment above explains why. Because it's a bug, it's missed in any spec or proposal, it's not documented, more other - code comments say that it should not be here, because the additional time of its availability could cause new cases of incorrect usage.

PR #990, opened 2021-09-30
Add `String.cooked` proposal

zloirock:
https://github.com/bathos/proposal-string-cooked

Stage 1 proposal.


zloirock:
Published.

Issue #969, opened 2021-08-07
structuredClone


zloirock:
[`structuredClone` method now exposed in the HTML spec.](https://github.com/whatwg/html/pull/3414) It's not a part of ECMAScript, but it's a very useful cross-platform feature, so makes sense to add it to `core-js`. If someone wanna contribute something significant to `core-js` - it's a good chance.

BasixKOR:
FYI, This is the algorithm used for structured clone [serialization](https://html.spec.whatwg.org/multipage/structured-data.html#structuredserializeinternal) and [deserialization](https://html.spec.whatwg.org/multipage/structured-data.html#structureddeserialize).

And can I get a chance to implement this? Not very familar with core-js codebase but I really wanna give it a try :)

zloirock:
@BasixKOR sure. Ask me if you will have any questions.

petamoriken:
I made a list of `Serializable`s and `Transferable`s from WebIDL. I hope this helps :)

https://gist.github.com/petamoriken/3802602b8e93d89e5b4c21e36683cadb#gistcomment-3899514

petamoriken:
The next release of Node.js (v17.0.0) will include `structuredClone` and `DOMException`.
https://github.com/nodejs/node/pull/40119

I think it would be nice to add support for `DOMException`.
related: https://github.com/zloirock/core-js/issues/938

zloirock:
@petamoriken thanks! However, I'm not sure that the implementation of the transferable behavior is possible in most cases.

BasixKOR:
The ongoing implementation of `structuredClone` uses DOMException if it's available, and fallbacks to plain Error if there isn't. I guess the support table should be updated to reflect the support status of Deno and Node.js.

zloirock:
https://github.com/whatwg/html/pull/5749

PR #991, opened 2021-10-03
Add `DOMException`

### LINKED ISSUE #938: Support DOMException constructor
It would be nice to have the DOMException constructor polyfilled for IE.

Found an imlementation in fetch.js https://github.com/github/fetch/blob/master/fetch.js#L487

zloirock:
Fixes #938

Could be useful for #969 / #984.

Includes also `Error#toString` polyfill.


packages/core-js/modules/web.dom-exception.js
```diff
@@ -0,0 +1,101 @@
...
+    type: DOM_EXCEPTION,
+    name: name,
+    message: message,
+    code: code
+  });
+  if (!DESCRIPTORS) {
+    this.name = name;
+    this.message = message;
+    this.code = code;
+  }
```
petamoriken, line 66: `DOMException` should have the `stack` property.

> Additionally, if an implementation gives native Error objects special powers or nonstandard properties (such as a stack property), it should also expose those on DOMException instances.

https://heycam.github.io/webidl/#es-DOMException-specialness

In Deno, the `stack` property is copied from `Error` object.
https://github.com/andreubotella/deno/blob/1debadecb6606c1304bf750bba9c9b360e5afafa/ext/web/01_dom_exception.js#L99-L104

NOTE: Current Chrome and Safari seem to violate the spec without the `stack` property when creating a `DOMException` from constructor 😅
https://wpt.fyi/results/webidl/ecmascript-binding/es-exceptions/DOMException-custom-bindings.any.html

zloirock: Thanks. It is still in WIP. In parallel, I'm working on the Error#cause polyfill and some other Error-related features. But the main problem is that the top LOC in the error stack will show the polyfill wrapper - here the same. I didn't add Error#cause polyfill for a long and waited when it will be supported by actual versions of Chrome, FF, Safari - it's enabled in Safari 15 that was published some days ago.

Another problem that should be solved - what should be done with incomplete implementations. Complete replacement could break the current code. In some cases, patching is hard, in other cases - impossible.

packages/core-js/modules/web.dom-exception.js
```diff
@@ -0,0 +1,150 @@
...
+  if (HAS_STACK) {
+    defineProperty(this, 'stack', createPropertyDescriptor(1, Error('DOMException: ' + message).stack));
```
petamoriken, line 71: Sorry to interrupt you while you are working on it, but perhaps you can use the following code.

```js
var error = Error(message);
error.name = DOM_EXCEPTION;
defineProperty(this, 'stack', createPropertyDescriptor(1, error.stack);
```

zloirock: Seems so, thanks.

zloirock:
Published.

PR #995, opened 2021-10-14
Add `Number.prototype.toExponential` polyfill

zloirock:
Fixes [those](https://github.com/kangax/compat-table/pull/1752) and some other bugs.


zloirock:
Published.

PR #1001, opened 2021-10-27
Changes from October 2021 TC39 meeting

zloirock:
https://github.com/babel/proposals/issues/77

- [x] Move `Array` grouping to stage 2
- [x] Add `Array.prototype.groupByMap`
- [x] Drop `@@species` support from `Array.prototype.groupBy`
- [x] Drop `%TypedArray%.prototype.groupBy`
- [x] Drop `%TypedArray%.fromAsync`

---

Separately:
- [x] `Error` `cause` https://github.com/zloirock/core-js/pull/993
- [x] `String.cooked` https://github.com/zloirock/core-js/pull/990
- [x] Change `Array` by copy https://github.com/zloirock/core-js/pull/923

PR #1004, opened 2021-11-03
feat: add array.groupByMap

Jack-Works:


Jack-Works:
That proposal also adds https://tc39.es/proposal-array-grouping/#sec-array.prototype.groupbymap but no `%TypedArray%.prototype.groupBy`

zloirock:
Thanks! It's already tracked here https://github.com/zloirock/core-js/pull/1001

Jack-Works:
I have removed TypedArray.groupBy

zloirock:
@Jack-Works it's a breaking change, so it can be removed only in `core-js@4`.

Jack-Works:
> @Jack-Works it's a breaking change, so it can be removed only in `core-js@4`.

I reverted that change. Do core-js provide semver constraints even for proposals? I thought proposals are unstable

zloirock:
Yes, sure. Breaking changes can be only in cases that *most likely* still no one used, like removing `@@species` support from `Array.prototype.groupBy` in #1001. Removing `esnext.typed-array.group-by` will break too many cases of code generated by `@babel/preset-env`.

zloirock:
Published.

PR #1012, opened 2021-11-22
Extension of the API

zloirock:
(sure, without breaking changes)

1. Adds a new `core-js/actual/**` namespace.

We will have 4 main namespaces:
- `/es/` that contains only stable ECMAScript features
- `/stable/` that contains stable ECMAScript features and features from web standards
- **`/actual/` that contains stable ECMAScript features, stage-3 ECMAScript proposals and features from web standards** that will be the recommended way of `core-js` usage
- `/features/` that contains only stable ECMAScript features

In `core-js@4`:
  - `import 'core-js'` will not contain early stage proposals and will be an equal of `import 'core-js/actual'`
  - `/features/` will be renamed to something more descriptive

2. Adds entry points for each finished post-ES6 proposal. For example, you will be `import 'core-js/proposals/string-padding'` for import [old string padding proposal](https://github.com/tc39/proposal-string-pad-start-end).

In `core-js@4`:
  - `core-js/stage/4` entry point will be removed since it contains proposals finished only after the latest `core-js` major release. For usage of them will be required to import it from `/es|stable|actual|features/` namespaces or by a proposal name, like in the example above.

---

**I'm not completely sure in the `/actual/` name, does anyone have better ideas?**


zloirock:
Published.

PR #1016, opened 2021-12-02
Add `Function.prototype.unThis` proposal

zloirock:
Stage 0.

https://github.com/js-choi/proposal-function-un-this


zloirock:
Published.

PR #1018, opened 2021-12-03
Add `Iterator.prototype.toAsync`

zloirock:
https://github.com/tc39/proposal-iterator-helpers/issues/160#issuecomment-984635413


zloirock:
Published.

PR #1019, opened 2021-12-07
Legacy `String#at` workaround

zloirock:
For a long time, modern `String#at` was not added because of the conflict with [an obsolete proposal](https://github.com/mathiasbynens/String.prototype.at).

However, even now loading `core-js` with stage-0 proposals can break modern `String#at`.

We can't just remove this proposal before `core-js@4` since it will break too many cases.

However, we can bring the obsolete `String#at` module semantics closer to the modern `String#at` for avoiding the most conflicts. It's mainly allowing negative indexes, but with saving multibyte semantic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Changelog
 ##### Unreleased
-- Nothing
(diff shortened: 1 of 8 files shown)


zloirock:
Published.

PR #1029, opened 2021-12-30
refactor: simplify expression.

zhangenming:


zloirock:
Thanks, but in this case, we lose check of the existence of the passed method. Something like this:
```js
uncurryThis(Array.prototype.groupBy); // => method or, if `groupBy` is not exist, `undefined`
```

zloirock:
I don't think that it's a simplification - for many, the first line will seem even more confusing and this does not affect performance, however, let it be.

Issue #1034, opened 2022-01-05
Is that `Array.prototype.concat` will be removed during importing polyfilll?


aleen42:
Recently, I have found that if I have shimmed `Function.prototype.bind` at first:

```js
var $A = Array.from = function (iterable) {
    if (!iterable) return [];
    if (iterable.toArray) {
        return iterable.toArray();
    } else {
        var results = [];
        for (var i = 0; i < iterable.length; i++)
            results.push(iterable[i]);
        return results;
    }
};

Function.prototype.bind = function () {
    var __method = this, args = $A(arguments), object = args.shift();
    return function () {
        return __method.apply(object, args.concat($A(arguments)));
    };
};
```

And then, when running in the process of importing `core-js/shim`, `Array.prototype.concat` has been removed, and an exception was thrown when `core-js` tries to call shimmed `Function.prototype.bind`.

I just want to confirm whether it is due to the commit (277cb333980e6c3f5fd37f8ed9867de272d027e9), which wants to protect some built-in methods?

zloirock:
> Is that `Array.prototype.concat` will be removed during importing polyfill?

Only in case if it should be polyfilled and immediately will be replaced to the proper method.

> And then, when running in the process of importing `core-js/shim`, `Array.prototype.concat` has been removed, and an exception was thrown when `core-js` tries to call shimmed `Function.prototype.bind`.

I can't reproduce that. I see only some possible issues from your incomplete `.bind` implementation that does not work with constructors.

zloirock:
For example, your code, `core-js` import and `core-js` test case:

![image](https://user-images.githubusercontent.com/2213682/148197181-46b90381-fcfa-4334-a1b6-6b4a16a7a860.png)

Issues only in your `.bind`.

aleen42:
In my case, I have caught an error thrown when accessing `args.concat` in my code. It seems wiered.

zloirock:
Please, add a reproducible example. I can't reproduce it just with this code.

aleen42:
@zloirock Denis, we can reproduce via my commit (https://github.com/aleen42/core-js/commit/937d46474547280a43f08b380ebc4adfb116dee8)

```
$ npm run test-pollution

> test-pollution
> karma start -f=pollution.js,packages/core-js-bundle/index.js,tests/bundles/tests.js

[...]

PR #1035, opened 2022-01-06
chore: add test cases to avoid internal usage of third-party `.bind` #1034

aleen42:


m0dul3:
d

zloirock:
Thanks, but I think that it's a too specific case, and adding it to the CI tests will just eat more resources than it could add benefits.

PR #1054, opened 2022-03-08
Add jsDelivr hits badge

MartinKolarik:
Hi, Martin from [jsDelivr CDN](https://www.jsdelivr.com/) here. I've noticed `core-js-bundle` recently made a big jump in popularity and become the most popular npm package we currently serve, with 15 billion downloads in the past month - I just wanted to say congrats! 🥳

<div align="center">

</div>

In case you'd like to showcase this number, we also have some stats badges similar to npm so I added one to the README - if not, feel free to close.


zloirock:
Great, thanks.

PR #1072, opened 2022-04-22
fix: issue on uid library after consecutive optimizations

denn1s:
Avoids the unusual syntax to get the Number.prototype.toString that might confuse code optimizers.

In our environment the line

`var toString = uncurryThis(1.0.toString);`

After optimizing the library during build, gets changed to 

`var toString = uncurryThis(1 .toString);`

Which is functionally identical, however our cloud provider once again minifies the code and removes the space

`var toString = uncurryThis(1.toString);`

Which is an invalid syntax. This pull requests tries to simplify the function to avoid this behavior on optimizers. This also helps in eliminating an import that was not needed.

### DIFF
--- a/packages/core-js/internals/uid.js
+++ b/packages/core-js/internals/uid.js
@@ -1,9 +1,7 @@
-var uncurryThis = require('../internals/function-uncurry-this');
-
 var id = 0;
 var postfix = Math.random();
-var toString = uncurryThis(1.0.toString);
 
 module.exports = function (key) {
-  return 'Symbol(' + (key === undefined ? '' : key) + ')_' + toString(++id + postfix, 36);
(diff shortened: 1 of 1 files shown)


zloirock:
Thanks for the PR, but I can't accept it. `Number.prototype.toString` can be patched in userland that makes internal behaviour of polyfills observable and breakable.

This is an issue for code optimizers that it can break - they should not change valid JS code to invalid.

denn1s:
I see, I understand the need to consider use space overriding, which is something I hadn't considered so I thought it was an easy fix. The optimization that removes the space actually occurs with cloudflare cache optimizing,  so I thought it would be good to patch as soon as possible. I'll close the pr.

PR #1075, opened 2022-05-01
Update usage.mjs

ImanSharaf:
Fixed a problem: Insecure file download over HTTP. In case of a man-in-the-middle attack, the attacker can change the data.


zloirock:
Thanks, however that no makes big sense. After this request, this script should check websites from the list by both protocols - `http` and `https`. However, let it be, I'll accept it.

PR #1090, opened 2022-06-09
Updating array grouping names.

hemanth:
[grouBy is now group](https://github.com/tc39/proposal-array-grouping#why-group-and-not-groupby)

//cc @jridgewell


zloirock (changes requested):
Thanks for the PR, however already exists https://github.com/zloirock/core-js/pull/1082 - most likely, it will be landed in the next week. In form from this PR, it can't be added - it's a breaking change - we can't allow it without a major `core-js` bump - old names should not be removed.

hemanth:
Ah, got it. Nods, it is a breaking change. Will close this PR on behalf of the other.

PR #1100, opened 2022-07-04
fix(*): Typo

dev-itsheng:


tests/pure/es.map.js
```diff
@@ -251,7 +251,7 @@ QUnit.test('Map#size', assert => {
     const getter = sizeDescriptor && sizeDescriptor.get;
     const setter = sizeDescriptor && sizeDescriptor.set;
     assert.same(typeof getter, 'function', 'size is getter');
-    assert.same(typeof setter, 'undefined', 'size isnt setter');
+    assert.same(typeof setter, 'undefined', 'size isn`t setter');
```
zloirock, line 254: Thanks. Please, use `'` here and in other cases.

dev-itsheng: Since I noticed a lot of similarities in the source code with backticks, I followed this practice for consistency with the existing code.

Well, next I try to modify everything I see.

zloirock: Ususally, I use `"` for strings if `'` should be used inside and `'` in all other case. Thanks!

dev-itsheng: In my opinion, it shouldn't be decided whether a string should be wrapped in double quotes or single quotes just because it contains single quotes.

[ESLint related documentation also describes](https://eslint.org/docs/latest/rules/quotes#js-toc-label): Many codebases require strings to be defined in a consistent manner.

Then I looked at some other source code, such as several loaders of webpack, they also used the backslash and single quotes.

zloirock: However, here use this coding style and it is [defined in the `eslint` config](https://github.com/zloirock/core-js/blob/7bbe8744efbcdc13bd23341c37d4aed58d3c9ad7/.eslintrc.js#L272-L273) - [see `avoidEscape` option](https://eslint.org/docs/latest/rules/quotes#avoidescape). It seems now ESLint options for this rule changed, so need to fix it.

PR #1095, opened 2022-06-21
Split `README` into files located at `/docs`

Rudxain:
# Current plan
Big and "medium" (relatively) sections from the `README.md` will be moved to the `/docs/` directory. Categories will have dedicated inner folders where files will be placed, other files will be placed directly at `docs/` if they are their own "category". To preserve as much external links as possible, the index will be the same (almost, because inner links will be moved to indexes at their respective files) and the sections will have headings that, when clicked, redirect to the corresponding file. This is the reason why the branch name is `readme-redirect`

## Links to more info
https://github.com/zloirock/core-js/pull/1095#issuecomment-1163496767
https://github.com/mdn/content/issues/17561#issuecomment-1164096932
https://github.com/zloirock/core-js/pull/1095#pullrequestreview-1026895683

# Directory tree
This section is about the planned structure of the `/docs` repo tree. This is subject to change and open to suggestions. Indentation = nesting depth. `# ` means it's a heading/section, not a file.
```
usage.md //should this be a folder?
	# Installation
	# `postinstall` message
	# CommonJS API
	# Babel
	# swc
	# Configurable level of aggressiveness
	# Custom build
compatibility-data.md //this includes "Supported engines"
features/
	ecmascript/
		object.md //prefixes seem useless in a folder
		function.md
		error.md
		array.md
		string and regexp.md //is this name good?
		number.md
		math.md
		date.md
		promise.md
		symbol.md
		collections.md
			# Map
			# Set
			# WeakMap
			# WeakSet
		typed-array.md
		reflect.md
		json.md
		global-this.md
	proposals/
		README.md //this contains the index and groups proposals by stage
		//each proposal is its own file, named in the same way as the files in `/packages/core-js/proposals/`
	web-standards/
		structured-clone.md
		base64-utility-methods.md
		set-timeout and set-interval.md
		set-immediate.md
		queue-microtask.md
		url.md
		dom-exception.md
		dom-collections.md
	iteration-helpers.md
missing-polyfills.md
```

# Related Issues
Fixes #1053

# Etc
Other stuff that this PR does is fix some typos and minify URLs

PR #1105, opened 2022-07-12
Update README.md

mooseki:
Add collapse/expand to ##Index


zloirock (commented):
Thanks, however, I don't think that it's a good idea. Anyway, it's not required because of #1095.

PR #1118, opened 2022-09-06
Add `self`

zloirock:
The [Minimum Common Web Platform API](https://common-min-api.proposal.wintercg.org/) implies the presence of `self` in all web-interoperable runtimes.

By the [spec](https://html.spec.whatwg.org/multipage/window-object.html#dom-self), it should be a getter.

image

However, some old and even actual environments add it as a usual property.

https://github.com/denoland/deno/issues/15765

Initially, I was skeptical about the compatibility of this with NodeJS, however, exploring GitHub shows that detection of missed `self` is usually used only in testing environments or in complex with other methods and `self` is polyfilled too often.

https://github.com/wintercg/proposal-common-minimum-api/issues/28

Thoughts?


zloirock:
More than a month has passed, but [I still don't see any feedback from The Minimum Common Web Platform API team](https://github.com/wintercg/proposal-common-minimum-api/issues/28). If I will not see any objections, it will be merged soon.

PR #1127, opened 2022-09-23
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,9 @@ name: ci
 
 on: [push, pull_request]
 
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
 jobs:
   lint:
     runs-on: ubuntu-latest


zloirock:
@sashashura you already added it https://github.com/zloirock/core-js/pull/1124, however, I temporarily reverted those changes since in the current form it breaks CI, see https://github.com/zloirock/core-js/actions/runs/3091251960

sashashura:
First of all sorry for the second pull request. I didn't mean that. Looking at the run I'm puzzled why it happened. The HTTP error is 429 (Too many requests), not forbidden. The code is quite standard - think there many workflows doing checkout in matrix. I wonder if this was unrelated... I see there were two runs. Did you start the second one manually?

sashashura:
I see there is another run that was failing in the same way with my change reverted: https://github.com/zloirock/core-js/actions/runs/3092396775 As you can see it wasn't related to permissions:
image

zloirock:
Here is the same https://github.com/zloirock/core-js/actions/runs/3092396775. I have no idea why it returns 429 error code - however, it's the result of changing permissions.

sashashura:
It is the same as I posted screenshot. As you can see at the top I have expanded the permissions it used to run. You can also click View Workflow file and see it wasn't with my changes. Anyway fell free to close the pr.
image

zloirock:
@sashashura it's the result of calling CI after pushing some commits together, including revert of permissions changes:
image
so it's called with the previous workflow.

zloirock:
Let's rerun https://github.com/zloirock/core-js/actions/runs/3092396775 and check it again.

zloirock:
Maybe it was an error on the GitHub API side. Let's try one more time.

Issue #1130, opened 2022-09-28
More `String.prototype.endsWith` problems on Nashorn


AprilArcus:
Now that https://github.com/zloirock/core-js/commit/27b563077e152ca1ef5d09e8f798e0077d6c1991 and https://github.com/zloirock/core-js/pull/1129 are merged, the original Nashorn error of `Java::JavaLang::AssertionError (unknown call type GET:METHOD|PROPERTY|ELEMENT:call(Object)Object@jdk.nashorn.internal.scripts.Script$Recompilation$3$\^eval\_/0x0000000800765040)` from https://github.com/zloirock/core-js/issues/1128 has been addressed, and the polyfill can be loaded successfully. However, when I try to use a function such as `endsWith`, I get this new error:

`Java::JavaxScript::ScriptException (TypeError: Can not invoke method [jdk.dynalink.beans.SimpleDynamicMethod boolean java.lang.String.endsWith(String)] with the passed arguments; they do not match any of its method signatures. in <eval> at line number 1)`

The problem originates with the fact that Nashorn implementation of `endsWith` can't be called with the `endPosition` parameter:

```
> engine.eval("Function.prototype.call.apply(''.endsWith, ['foo',
 'o'])")
< true
engine.eval("Function.prototype.call.apply(''.endsWith, ['foo', 'o', 3])")
Java::JavaxScript::ScriptException
```

In order to correct this, the return value from uncurryThis should be falsey here:

https://github.com/zloirock/core-js/blob/master/packages/core-js/modules/es.string.ends-with.js#L13

Then the native implementation will be bypassed during this check:

https://github.com/zloirock/core-js/blob/master/packages/core-js/modules/es.string.ends-with.js#L34

PR #1131, opened 2022-09-28
Fix more Nashorn problems

AprilArcus:
Fixes https://github.com/zloirock/core-js/issues/1130

### DIFF
--- a/packages/core-js/internals/function-uncurry-this.js
+++ b/packages/core-js/internals/function-uncurry-this.js
@@ -6,8 +6,14 @@ var call = FunctionPrototype.call;
 var uncurryThis = NATIVE_BIND && bind.bind(call, call);
 
 module.exports = function (fn) {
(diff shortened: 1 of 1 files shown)


packages/core-js/internals/function-uncurry-this.js
```diff
@@ -6,8 +6,12 @@ var call = FunctionPrototype.call;
 var uncurryThis = NATIVE_BIND && bind.bind(call, call);
 
 module.exports = function (fn) {
+  var isNativeFunction = fn instanceof Function;
   // Nashorn bug, https://github.com/zloirock/core-js/issues/1128
-  return fn && (NATIVE_BIND && fn instanceof Function ? uncurryThis(fn) : function () {
-    return call.apply(fn, arguments);
-  });
+  if (!isNativeFunction) return false;
```
zloirock, line 11:
```suggestion
  if (!isNativeFunction) return;
```

PR #1150, opened 2022-11-29
Update `Set` methods proposal

zloirock:
- [`Set` methods proposal](https://github.com/tc39/proposal-set-methods):
  - Methods:
    - `Set.prototype.intersection`
    - `Set.prototype.union`
    - `Set.prototype.difference`
    - `Set.prototype.symmetricDifference`
    - `Set.prototype.isSubsetOf`
    - `Set.prototype.isSupersetOf`
    - `Set.prototype.isDisjointFrom`
  - Moved to Stage 3, [November 2022 TC39 meeting](https://github.com/babel/proposals/issues/85#issuecomment-1332175557)
  - Reimplemented with [new semantics](https://tc39.es/proposal-set-methods/):
    - Optimized performance (iteration over lowest set)
    - Accepted only `Set`-like objects as an argument, not all iterables
    - Accepted only `Set`s as `this`, no `@@species` support, and other minor changes
  - Added `/actual/` entries, unconditional forced replacement changed to feature detection
  - For avoiding breaking changes:
    - New versions of methods are implemented as new modules and available in new entries or entries where old versions of methods were not available before (like `/actual/` namespace)
    - In entries where they were available before (like `/full/` namespace), those methods are available with fallbacks to old semantics (in addition to `Set`-like, they accept iterable objects). This behavior will be removed from the next major release

cc @bakkot


packages/core-js/internals/set-intersection.js
```diff
@@ -0,0 +1,26 @@
...
+module.exports = function intersection(other) {
+  var O = aSet(this);
+  var otherRec = getSetRecord(other);
+  var result = new Set();
+  if (size(O) <= otherRec.size) forEach(O, function (e) {
+    if (otherRec.includes(e)) add(result, e);
+  });
+  else iterateSimple(otherRec.getIterator(), function (e) {
+    if (has(O, e)) add(result, e);
+  });
```
ljharb, line 24: this doesn't seem like it ensures the order in `result` of items contained in `O` match their ordering in `O` - iow, `new Set([1, 2, 3]).intersection(new Set([3, 2]))` should result in `new Set([2, 3])`, but i think this algorithm results in `new Set([3, 2])`.

I'm trying to think of a way to achieve it that doesn't involve iterating over the entirety of `O` - I'd love to see what you come up with.

zloirock: Thanks, I saw the problem with tests of this method, but you showed that the problem is in the missed 7e step.

It seems that without adding iteration over `O` ensuring `O.[[SetData]]` order is impossible.

ljharb: That was my conclusion as well. Perhaps @bakkot has a better idea?

bakkot: I don't think it's possible to do efficiently in userland. It's possible in engines only because they have access to the actual underlying data structure.

Issue #1154, opened 2022-12-05
Pure version of Number constructor


trosos:
Hi,
Is there a technical or philosophical reason not to have a pure version of the `Number` constructor?

I would appreciate it for transpiling things like `Number("0b11")` (using babel-plugin-transform-runtime).

Thank you for a great project!

zloirock:
Hi. It's a technical problem.

It's not a problem to add a pure polyfill for it, it's a problem to make it work with `babel-plugin-transform-runtime` as expected in all cases.

Initially, it was added, but it broke some projects.

For example, someone checks a type like this:

```js
const number = 1;

number.constructor === Number; // => true
```

But with runtime, we could have:

```js
const _Number = require(/* something */);

const number = 1;

number.constructor === _Number; // => false
```

and vice versa.

It can work as expected only in the case of patching built-in prototypes, but it contradicts the pure nature.

PR #1155, opened 2022-12-07
Add pure version of the `Number` constructor

trosos:
Implements #1154


packages/core-js/modules/es.number.constructor.js
```diff
@@ -62,7 +64,8 @@ var toNumber = function (argument) {
 
 // `Number` constructor
 // https://tc39.es/ecma262/#sec-number-constructor
-if (isForced(NUMBER, !NativeNumber(' 0o1') || !NativeNumber('0b1') || NativeNumber('+0x1'))) {
+var shouldForce = !NativeNumber(' 0o1') || !NativeNumber('0b1') || NativeNumber('+0x1');
+if (IS_PURE || isForced(NUMBER, shouldForce)) {
```
zloirock, line 68: In the case of the pure version, in environments that support the required `Number` constructor behavior, complex string parsing should not be used.

trosos: You are right. This was my first thought. I just got confused by the fact that the node unit tests then run on the native builtin, rather than on the polyfill.

trosos: I believe that I have addressed your concern in my last commit. Is there anything other I can do about this PR?
Or is my approach just plain wrong that you don't want to waste your time trying to correct it? :-)

zloirock: Hey, give me some time - I also have other tasks -) Now I see that this is not a completely correct approach, I'll explain the details a little later.

trosos: You have as much time as you want. Didn't mean to be bratty.

zloirock: With the second your commit, in case of correct native `Number` behavior, in the pure version would be exported native `Number` - that means native `Number` pollution on polyfilling static `Number` properties. Also, there were some issues with reexport of statics.

For cases like this, the `export` helper has a `wrap` option.

However, in the pure version, it does not ensure proper `.name`, `.length`, and (only in some specific cases like this) subclassing - it's accepted limitations of the pure version.

trosos: Of course, such a silly mistake, I'm sorry about that.
I totally forgot about the static properties :(

PR #1159, opened 2023-01-03
Add typescript information for core-js-builder

susnux:
Add declarations for `core-js-builder` so it can be used from within TypeScript projects.


zloirock:
Thanks, that makes sense.

packages/core-js-builder/index.d.ts
```diff
@@ -0,0 +1,29 @@
+type Summary = {
+    size: boolean,
+    modules: boolean
+};
+
+type Options = {
+    /** entry / module / namespace / an array of them, by default - all `core-js` modules */
+    modules?: string[],
```
zloirock, line 8: It can be a string, regex, or an array of them, see https://github.com/zloirock/core-js/tree/master/packages/core-js-compat

packages/core-js-builder/index.d.ts
```diff
@@ -0,0 +1,29 @@
...
+    /** a blacklist of entries / modules / namespaces, by default - empty list */
+    exclude?: string[],
```
zloirock, line 10: It can be a string, regex, or an array of them.

packages/core-js-builder/index.d.ts
```diff
@@ -0,0 +1,29 @@
...
+    /** optional browserslist or core-js-compat format query */
+    targets?: string | string[],
```
zloirock, line 12: Also, it can be an object.

packages/core-js-builder/index.d.ts
susnux: Thank you, resolved :)

packages/core-js-builder/index.d.ts
susnux: Thank you, resolved :)

susnux:
Thank you for the feedback @zloirock ! I resolved the review notes and added the missing types.

packages/core-js-builder/index.d.ts
```diff
@@ -0,0 +1,33 @@
+type Summary = {
+    size: boolean,
+    modules: boolean
+};
+
+type StringOrRegExp = string | RegExp;
+
+type MinEnvTarget = Record<string, string> & { esmodules?: boolean };
```
zloirock, line 8: It's a little more complex. A record can contain string, boolean, or array values. See, for example, Babel typing in a similar case https://github.com/babel/babel/blob/d3bce24473c5efef1a80746bdb7ee615bb29f778/packages/babel-helper-compilation-targets/src/types.d.ts

susnux: For array I could find a reference within the source code (fixed this!), but I only see one boolean, the `esmodules` option.

zloirock: Yes, and `esmodules` can be mixed with other targets, and that's not allowed with the current typing.

susnux: Should word now with `Record<string, string | readonly string[]> | { esmodules?: boolean };`

packages/core-js-builder/README.md
```diff
@@ -32,3 +32,5 @@ const bundle = await builder({
   filename: PATH_TO_MY_COREJS_BUNDLE,
 });
 ```
+
+ℹ️ When using TypeScript, make sure to set `esModuleInterop` to `true`.
```
zloirock, line 36: I don't often use TypeScript, but IIRC with `export = ...` `esModuleInterop: true` is not required, or I missed something?

packages/core-js-builder/index.d.ts
```diff
@@ -0,0 +1,33 @@
+type Summary = {
+    size: boolean,
+    modules: boolean
```
zloirock, line 3: Those `Summary` fields are optional. `Summary` also can be a boolean.

packages/core-js-builder/README.md
zloirock: It seems so:
image

zloirock:
I made it stricter and fixed some issues.

susnux:
> I made it stricter and fixed some issues.

Have you removed the doc comments on purpose? I aded them to have information on what each option does within your IDE

zloirock:
I'm not sure that it's something useful, but if you want - ok, let's return them.

packages/core-js-builder/README.md
susnux: If you omit that setting you get following error:
```
9 import builder from "core-js-builder"
         ~~~~~~~

  ../core-js/packages/core-js-builder/index.d.ts:60:1
    60 export = builder;
       ~~~~~~~~~~~~~~~~~
    This module is declared with 'export =', and can only be used with a default import when using the 'allowSyntheticDefaultImports' flag.
```

If you add `allowSyntheticDefaultImports` it compiles, but the compiled output is invalid as it looks like this (note the `.default`):
```js
const core_js_builder_1 = require("core-js-builder");
// ...
const bundle = await (0, core_js_builder_1.default)(this.options);
```

If you change the import to `import * as builder from 'core-js-builder'` the output is correct, so no `.default` is added, but you get following error:
```
src/plugin.ts:9:26 - error TS2497: This module can only be referenced with ECMAScript imports/exports by turning on the 'esModuleInterop' flag and referencing its default export.

9 import * as builder from "core-js-builder"
                           ~~~~~~~~~~~~~~~~~
```
---
So the only way to import the module without errors is setting `esModuleInterop` to `true` and referencing the default export.
(I use TypeScript 4.9.4)

susnux:
> I'm not sure that it's something useful, but if you want - ok, let's return them.

I like explanations within the IDE, so you do not have to switch to your browser for the example within the README :smile: 

| without | with |
|-------------|-------|
| without comment hovering an option |hovering an option with comment |

susnux:
Any reason why you closed the PR?

zloirock:
Sorry, I mistakenly forced pushed on rebase - and it seems since because of this PR is closed, I have no longer access to this branch. I'll move commits from this PR to a new branch.

zloirock:
#1163

PR #1190, opened 2023-02-15
Omit victim-blaming language in original letter

Gordon01:
This pull request aims to improve the clarity and ethicality of the original letter by omitting the victim-blaming language used by the author. The original letter contains language that attempts to shift blame from the driver to the pedestrians who were hit, which is unacceptable and harmful.

The goal of this pull request is to provide a revised version of the letter that accurately conveys the events that occurred without perpetuating harmful stereotypes or blaming the victims. The revised letter focuses on the driver's responsibility in the accident and the consequences of their actions, rather than on the behavior of the pedestrians.

By making these changes, we can create a more responsible and compassionate narrative that respects the dignity and agency of all individuals involved in the accident. It is important to recognize the devastating impact that car accidents can have on victims and their families, and to take responsibility for our actions as drivers.

Overall, this pull request seeks to promote a more ethical and respectful approach to discussing car accidents, one that recognizes the responsibility of drivers and the impact of their actions on others.


myceera:
Wonderful PR! As a transport planner I can see the effort and thoughtfulness put into making these changes, and the result is a much more appropriate and respectful representation of the situation.

This new version takes a more ethical approach to discussing car accidents which actually not ‘accidents’ and might be prevented not by demanding legal obligations for pedestrians to wear bright clothes, but with more thoughtful transport planning.

This PR acknowledges the responsibility of drivers and the impact of their actions on others. It provides a more accurate and honest account of the accident without trying to shift blame onto the victims or minimize the seriousness of the situation.

PR #1225, opened 2023-03-18
update some judgement condition

mek1986:
1.== to ===
2.!= to !==


zloirock:
I thought about [disallowing](https://eslint.org/docs/latest/rules/eqeqeq) `==` / `!=` for a long time, however, in the current `core-js` style it's allowed **in safe cases**. Any argumentation that in those cases should be used strict comparison?

zloirock:
I enabled `eqeqeq`.

mek1986:
> I enabled `eqeqeq`.
06667ADD

PR #1235, opened 2023-04-08
Add typescript declarations for core-js-compat

susnux:
Reuse it for the core-js-builder declarations to prevent duplicated code


zloirock:
Thanks! However, `core-js-compat` contains some more entries.

susnux:
@zloirock I have added more type declarations, but do you think type declarations for the `helpers` are required as well?

zloirock:
> (CI fails because of a NodeJS bug [nodejs/node#47505](https://github.com/nodejs/node/issues/47505), I'll add a workaround soon.)

Added a workaround in `master`.

susnux:
@zloirock 
> Added a workaround in `master`.

Rebased works now.

PR #1241, opened 2023-04-28
Add guard for ActiveXObject

chrbala:
My org has a build artifact linting system which gets tripped up by the try/catch block around Object.create. This would be solved for us with an `if` statement as proposed in this PR. Is it possible to get this change added?


zloirock:
Adding `if` block does not change anything here for the runtime - if global `ActiveXObject` is not declared, the error will be thrown in `if (ActiveXObject)`.

It's only one of the hundreds of similar cases in `core-js`, so it's strange to prevent throwing an error only here.

It's rather the issue of your listing config.

PR #1242, opened 2023-05-01
Fix `Symbol` compatibility detection under Terser

laverdet:
The `String` constructor has a special case for `Symbol` conversions: https://tc39.es/ecma262/#sec-string-constructor-string-value

    a. If NewTarget is undefined and value is a Symbol, return
       SymbolDescriptiveString(value).
    b. Let s be ? ToString(value).

Therefore, the only difference between `String(value)` and `value + ''` is that the latter will throw for symbols.

Terser uses this to advantage by replacing invocations of `String(...)` with `... + ''`. Terser, to its credit, has this behavior gated behind the `unsafe` flag with the following documentation: "It enables some transformations that might break code logic in certain contrived cases, but should be fine for most code." For this particular code it is *not* fine though.

Under WebPack, Terser is typically run per chunk rather than per file. It is difficult or impossible to disable these optimizations for only core-js, only for the entire bundle. This optimization causes the Symbol constructor detection pathway to fail and fallback to string keys. After everything is said and done we end up with a confusing error message of "[object Generator] is not iterable".

The fix is simple, we simply bailout of this optimization in this critical path.

### DIFF
--- a/packages/core-js/internals/symbol-constructor-detection.js
+++ b/packages/core-js/internals/symbol-constructor-detection.js
@@ -1,7 +1,9 @@
 /* eslint-disable es/no-symbol -- required for testing */
 var V8_VERSION = require('../internals/engine-v8-version');
 var fails = require('../internals/fails');
-var $String = String;
+var global = require('../internals/global');
+
+var $String = global.String;
 
 // eslint-disable-next-line es/no-object-getownpropertysymbols -- required for testing
 module.exports = !!Object.getOwnPropertySymbols && !fails(function () {
--- a/packages/core-js/internals/symbol-constructor-detection.js
+++ b/packages/core-js/internals/symbol-constructor-detection.js
@@ -1,13 +1,16 @@
 /* eslint-disable es/no-symbol -- required for testing */
 var V8_VERSION = require('../internals/engine-v8-version');
 var fails = require('../internals/fails');
+var $String = String;
 
 // eslint-disable-next-line es/no-object-getownpropertysymbols -- required for testing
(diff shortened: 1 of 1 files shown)


zloirock:
It's not the only difference. For example,

```js
const value = { toString(){ return 'toString' }, valueOf() { return 'valueOf' } }
String(value) // => toString
value + ''    // => valueOf
```

This is too dangerous a transform.

However, I agree that it makes sense to accept this workaround.

zloirock (approved):

laverdet:
Sorry about the lint error!

packages/core-js/internals/symbol-constructor-detection.js
```diff
@@ -1,13 +1,16 @@
 /* eslint-disable es/no-symbol -- required for testing */
 var V8_VERSION = require('../internals/engine-v8-version');
 var fails = require('../internals/fails');
+var $String = String;
```
zloirock, line 4: In one moment, Terser could become smarter and "optimize" it even in this case. It's better to get `String` like here https://github.com/zloirock/core-js/blob/v3.30.1/packages/core-js/internals/regexp-sticky-helpers.js#L2-L5

ehoogeveen-medweb:
This might also help people still using [babel-preset-minify](https://babeljs.io/docs/babel-preset-minify) as the [minify-type-constructors](https://babeljs.io/docs/babel-plugin-minify-type-constructors) plugin (which is on by default) also performs this transform (by default). Though personally I switched due to other bugs (and general lack of maintenance).

PR #1279, opened 2023-08-02
Fix IEEE754 rounding rule from truncation to roundTiesToEven

petamoriken:
The decimal part of mantissa variable is truncated by bitwise operations, but it should be rounded (to even).

related https://github.com/petamoriken/float16/issues/952


zloirock:
Thanks -)

zloirock (approved):

PR #1284, opened 2023-08-22
Add `clearTimeout`

philipmw:
The functionality already existed and was tested by unit tests. This change exports it and clarifies misleading descriptions of its unit tests.


zloirock (changes requested):
Thanks for the PR, however, it makes no sense. `core-js` does not implement `setTimeout`, just fixes incorrect implementations, and does not touch `clearTimeout` at all - the global method should work fine.

You could leave the fixed name in the test and I'll accept this part of the PR.

philipmw:
Thanks for the feedback. I must have misunderstood what the code is doing. I'll close this request.

Issue #1286, opened 2023-09-01
An unexpected behaviour in strict mode in PhantomJS


aleen42:
Recently, I have found that when I upgraded `core-js`, a test case run in PhantomJS had failed with the following exception:

```
PhantomJS 2.1.1 (Mac OS 0.0.0) LOG: 'polyfill load failed', ReferenceError: Can't find variable: SAFE_CLOSING
exports
eval code
eval@[native code]
./node_modules/core-js/modules/es.array.from.js@http://localhost:9876/base/build/coremail/__karma__bundle/commons.js?407226f1f29b0c48cbb273c58df7aaa58a0e970e:3027:5
__webpack_require__@http://localhost:9876/base/build/coremail/__karma__bundle/runtime.js?3c2a1c270c4c70b6d7720b608d8bc64728c3bd1b:31:46
```

To trace the source, I have found that the behaviour existed due to the commit 11f4ec8bb27f0f5513b6c61a68b8623c64f5a1b2, which  specified `check-correctness-of-iteration.js` to execute in strict mode.

```js
'use strict';
var wellKnownSymbol = require('../internals/well-known-symbol');

var ITERATOR = wellKnownSymbol('iterator');
var SAFE_CLOSING = false;

try {
  var called = 0;
  var iteratorWithReturn = {
    next: function () {
      return { done: !!called++ };
    },
    'return': function () {
      SAFE_CLOSING = true;
    }
  };
  iteratorWithReturn[ITERATOR] = function () {
    return this;
  };
  // eslint-disable-next-line es/no-array-from, no-throw-literal -- required for testing
  Array.from(iteratorWithReturn, function () { throw 2; });
} catch (error) { /* empty */ }

module.exports = function (exec, SKIP_CLOSING) {
  if (!SKIP_CLOSING && !SAFE_CLOSING) return false; // Can't find variable: SAFE_CLOSING
  // ...
}
```

After a simple trial, I found that the execution of the following snippet has broken the scope:

```js
Array.from(iteratorWithReturn, function () { throw 2; });
```

PR #1287, opened 2023-09-01
handle missing scopes in strict mode when checking iteration #1286

aleen42:
related issue: #1286

### DIFF
--- a/packages/core-js/internals/check-correctness-of-iteration.js
+++ b/packages/core-js/internals/check-correctness-of-iteration.js
@@ -22,7 +22,9 @@ try {
 } catch (error) { /* empty */ }
 
 module.exports = function (exec, SKIP_CLOSING) {
-  if (!SKIP_CLOSING && !SAFE_CLOSING) return false;
+  try {
+    if (!SKIP_CLOSING && !SAFE_CLOSING) return false;
+  } catch (error) { return false; } // it is not safe when the `SAFE_CLOSING` is not in the current scope
   var ITERATION_SUPPORT = false;
   try {
     var object = {};


alinzaka (approved):

zloirock:
I know about some specific behavior in PhantomJS's strict mode, but that looks strange since this variable is defined in the upper scope. Moreover, `core-js` is tested in PhantomJS and tests passed without any problems. As an option, it can be related, for example, to a minifier. Are you sure that it's a fix for your case?

aleen42:
@zloirock It does cannot reproduce unless `eval`:

```js
"use strict";
eval("\nvar wellKnownSymbol = __webpack_require__(/*! ../internals/well-known-symbol */ \"./node_modules/core-js/internals/well-known-symbol.js\");\n\nvar ITERATOR = wellKnownSymbol('iterator');\nvar SAFE_CLOSING = false;\n\ntry {\n  var called = 0;\n  var iteratorWithReturn = {\n    next: function () {\n      return { done: !!called++ };\n    },\n    'return': function () {\n      SAFE_CLOSING = true;\n    }\n  };\n  iteratorWithReturn[ITERATOR] = function () {\n    return this;\n  };\n  // eslint-disable-next-line es/no-array-from, no-throw-literal -- required for testing\n  Array.from(iteratorWithReturn, function () { throw 2; });\n} catch (error) { /* empty */ }\n\nmodule.exports = function (exec, SKIP_CLOSING) {\n  if (!SKIP_CLOSING && !SAFE_CLOSING) return false;\n  var ITERATION_SUPPORT = false;\n  try {\n    var object = {};\n    object[ITERATOR] = function () {\n      return {\n        next: function () {\n          return { done: ITERATION_SUPPORT = true };\n        }\n      };\n    };\n    exec(object);\n  } catch (error) { /* empty */ }\n  return ITERATION_SUPPORT;\n};\n\n\n//# sourceURL=webpack://./node_modules/core-js/internals/check-correctness-of-iteration.js?");
```

packages/core-js/internals/check-correctness-of-iteration.js
```diff
@@ -22,7 +22,9 @@ try {
 } catch (error) { /* empty */ }
 
 module.exports = function (exec, SKIP_CLOSING) {
-  if (!SKIP_CLOSING && !SAFE_CLOSING) return false;
+  try {
+    if (!SKIP_CLOSING && !SAFE_CLOSING) return false;
+  } catch (error) { return false; } // it is not safe when the `SAFE_CLOSING` is not in the current scope
```
zloirock, line 27:
```suggestion
  } catch (error) { return false; } // workaround of old WebKit + `eval` bug
```

zloirock (approved):

PR #1294, opened 2023-09-30
Re-introduced `RegExp` escaping stage 2 proposal

zloirock:
https://github.com/tc39/proposal-regex-escaping

Includes a fix for https://github.com/tc39/proposal-regex-escaping/issues/62


MrPiNkY2022 (commented):
Great Work

PR #1302, opened 2023-11-11
Fix esnext.map.delete-all.js

jhmaster2000:
Currently the `Map#deleteAll` polyfill is returning a boolean whether it deleted everything successfully or not, this is not the correct behavior of the current ES proposal: https://tc39.es/proposal-collection-methods/#Map.prototype.deleteAll

The expected behavior is to return the map itself (`this`), so I've also removed all the successful deletion tracking code as it was no longer needed.

### DIFF
--- a/packages/core-js/modules/esnext.map.delete-all.js
+++ b/packages/core-js/modules/esnext.map.delete-all.js
@@ -8,11 +8,9 @@ var remove = require('../internals/map-helpers').remove;
 $({ target: 'Map', proto: true, real: true, forced: true }, {
   deleteAll: function deleteAll(/* ...elements */) {
     var collection = aMap(this);
-    var allDeleted = true;
-    var wasDeleted;
     for (var k = 0, len = arguments.length; k < len; k++) {
-      wasDeleted = remove(collection, arguments[k]);
-      allDeleted = allDeleted && wasDeleted;
-    } return !!allDeleted;
+      remove(collection, arguments[k]);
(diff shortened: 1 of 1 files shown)


zloirock:
Thanks, it's a good catch. However, let's take a look at [`Set#deleteAll` draft text](https://tc39.es/proposal-collection-methods/#Set.prototype.remove-elements) - that returns a boolean, but they should be consistent. It seems it's need to raise this issue in the proposal repo.

zloirock:
https://github.com/tc39/proposal-collection-methods/issues/2

zloirock:
See this issue https://github.com/tc39/proposal-collection-methods/issues/53 - it seems the current semantic is correct and it should be fixed in the spec text.

jhmaster2000:
Oh I missed that inconsistency, I did find it odd to return `this` as the boolean return does feel way more helpful too. I'll leave it up to you if you want to close this already or wait until there's an official resolution on the proposal issue, either works for me.

zloirock:
> I did find it odd to return `this` as the boolean return does feel way more helpful too.

If you think so, you can speak out in the above-mentioned issue -)

zloirock:
I don't think that we will have the answer to https://github.com/tc39/proposal-collection-methods/issues/53  soon, but it seems we can consider  https://github.com/tc39/proposal-collection-methods/issues/2 as that, so I think that we can close this PR.

Issue #1322, opened 2024-01-25
Spec Mismatch in `Array.of`


minseok-choe:
Following code behaves differently depending on whether the native version is loaded, or if the polyfill is required and loaded:
```javascript
let of = require("core-js-pure/es/array/of");

let handler = {
  has: () => {
    throw 42;
  },
};

let x = function () {
  return new Proxy({}, handler);
};

of.call(x, 0)
```

It throws with the polyfill, while it doesn't with the native implementation.
Although it is unusual to implement proxy to throw an exception, but even without throwing, it can still introduce silent bugs due to side effects.

zloirock:
From #1324:

> You are right, it's an optimization. Without this optimization, some methods worked unacceptably slow. However, now most methods where it's used are available in most engines, and `Proxy` is enough popular. I will think make it sense to accept it or not.

PR #1324, opened 2024-01-25
Fix create-property.js

minseok-choe:
This PR is for https://github.com/zloirock/core-js/issues/1322

The changes are as follows

```js
module.exports = function (object, key, value) {
  var propertyKey = toPropertyKey(key);
  if (propertyKey in object) definePropertyModule.f(object, propertyKey, createPropertyDescriptor(0, value));
  else object[propertyKey] = value;
};
```

```js
module.exports = function (object, key, value) {
  var propertyKey = toPropertyKey(key);
  definePropertyModule.f(object, propertyKey, createPropertyDescriptor(0, value));
};
```

I am not certain about the impact of this patch on performance - in case the existing code was intentionally written for performance benefits like JIT optimization, please handle it appropriately at your discretion.

### DIFF
--- a/packages/core-js/internals/create-property.js
+++ b/packages/core-js/internals/create-property.js
@@ -5,6 +5,5 @@ var createPropertyDescriptor = require('../internals/create-property-descriptor'
 
 module.exports = function (object, key, value) {
   var propertyKey = toPropertyKey(key);
-  if (propertyKey in object) definePropertyModule.f(object, propertyKey, createPropertyDescriptor(0, value));
-  else object[propertyKey] = value;
+  definePropertyModule.f(object, propertyKey, createPropertyDescriptor(0, value));
 };


zloirock:
This PR should be to `master` branch.

You are right, it's an optimization. Without this optimization, some methods worked unacceptably slow. However, now most methods where it's used are available in most engines, and `Proxy` is enough popular. I will think make it sense to accept it or not.

minseok-choe:
Sorry for the confusion. I changed the base branch to `master`.

minseok-choe:
Thank you for your hard work for maintaining this project. May I ask a question?

Does the "unacceptable slowness" come from the fact that `Object.defineProperty(object, key, createPropertyDescriptor(0, value))` cannot be optimized by engine, whereas `object[key] = value` can?
Or does it only occur when polyfill version is used for `definePropertyModule.f`?

zloirock:
> Does the "unacceptable slowness" come from the fact that `Object.defineProperty(object, key, createPropertyDescriptor(0, value))` cannot be optimized by engine, whereas `object[key] = value` can?

Yes. Some years ago it was dozens of times slower even in modern engines. I remember issues about optimizing this case in bug trackers of some engines. I didn't check the performance recently, so I don't know how it works in actual versions.

Issue #1327, opened 2024-02-05
Spec mismatch in `Array.prototype.reduce`


minseok-choe:
Following code behaves differently depending on whether the native version is loaded, or if the polyfill is required and loaded:

```javascript
const reduce = require("core-js-pure/actual/array/reduce");

const handler = {
  has: () => {
    throw 42;
  },
};

const y = new Proxy([], handler);

reduce(y, (e) => e);
```

PR #1328, opened 2024-02-05
Add missed check in `array-reduce.js`

minseok-choe:
This PR is about fixing [issue/1327](https://github.com/zloirock/core-js/issues/1327)

Missed check for [line 4 of spec.](https://262.ecma-international.org/14.0/#sec-array.prototype.reduce) is added


zloirock (approved):
Thanks -)

zloirock:
BTW it would be useful if you add tests for problem cases in the scope of the PR next time. However, it's not a critical requirement -)

PR #1333, opened 2024-02-23
Fix `object-key-internal.js`

stonechoe:
Following code behaves differently depending on whether the native version is loaded, or if the polyfill is required and loaded:

```js
var f = require("core-js-pure/es/object/keys");

var a = function () {};
Object.defineProperty(a, "valueOf", {value: 42, enumerable: false });

f(a).length === 0 // This should be true, but it is === 1 in polyfill
```

Current implementation of `object-key-internal.js` specially handles some buggy keys such as `toString` or `valueOf`, but it doesn't check if those keys are really enumerable, which is fixed in this PR.

### DIFF
--- a/packages/core-js/internals/object-keys-internal.js
+++ b/packages/core-js/internals/object-keys-internal.js
@@ -4,6 +4,8 @@ var hasOwn = require('../internals/has-own-property');
 var toIndexedObject = require('../internals/to-indexed-object');
 var indexOf = require('../internals/array-includes').indexOf;
 var hiddenKeys = require('../internals/hidden-keys');
+var call = require('../internals/function-call');
+var propertyIsEnumerableModule = require('../internals/object-property-is-enumerable');
 
 var push = uncurryThis([].push);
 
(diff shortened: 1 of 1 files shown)


minseok-choe:
Test failed, but it seems that an error occurred due to other factors during the process, preventing the tests from being executed at all. Could you please review this, if possible? Thank you. 😊

zloirock (commented):
This polyfill is oriented to ES3 engines where you can't define a non-enumerable property - all actual for `core-js` engines that allow a definition of non-enumerable properties already have `Object.keys`. IIRC initially this check was here, but it was removed for the described reason.

However, it could be useful in the case of handling prototypes. Anyway, it's not complete since this helper is also used in `Object.getOwnPropertyNames`.

I should think about it.

stonechoe:
I didn't know that this polyfill is oriented towards ES3 engines when I made this PR. Given your explanation, it seems appropriate to close this PR; Please handle it as you see fit.

zloirock:
Since anyway this logic will be removed from v4, let's close it.

PR #1346, opened 2024-05-30
update http to https(url-constructor-detection)

### LINKED ISSUE #1345: Can we upgrade http to https in url-constructor-detection.js ?
https://github.com/zloirock/core-js/blob/f1303f38bd80474f950cadb3e802db1c2618a2c5/packages/core-js/internals/url-constructor-detection.js#L11

Because core.js, as a dependency of many underlying compilation packages, will appear in chunk-vendors after packaging, but some projects cannot require http, so the packaged code scan will scan out http in url-constructor-detection.js.

http://a
http://a/c%20d?a=1&c=3
...

wChenonly:
fix #1345

### DIFF
--- a/packages/core-js/internals/url-constructor-detection.js
+++ b/packages/core-js/internals/url-constructor-detection.js
@@ -8,7 +8,7 @@ var ITERATOR = wellKnownSymbol('iterator');
 
 module.exports = !fails(function () {
   // eslint-disable-next-line unicorn/relative-url-style -- required for testing
-  var url = new URL('b?a=1&b=2&c=3', 'http://a');
+  var url = new URL('b?a=1&b=2&c=3', 'https://a');
   var params = url.searchParams;
   var params2 = new URLSearchParams('a=1&a=2&b=3');
   var result = '';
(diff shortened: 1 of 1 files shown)


packages/core-js/internals/url-constructor-detection.js
```diff
@@ -24,19 +24,19 @@ module.exports = !fails(function () {
...
+    || new URL('https://тест').host !== 'xn--e1aybc'
     // not escaped in Chrome 62-
-    || new URL('http://a#б').hash !== '#%D0%B1'
+    || new URL('https://a#б').hash !== '#%D0%B1'
     // fails in Chrome 66-
     || result !== 'a1c3'
     // throws in Safari
-    || new URL('http://x', undefined).host !== 'x';
+    || new URL('https://x', undefined).host !== 'x';
 });
```
zloirock, line 42: That also should be done in https://github.com/zloirock/core-js/blob/master/tests/compat/tests.js#L94

wChenonly: ok

PR #1368, opened 2024-09-20
Fix to avoid minification problem with esbuild

diegovdev:
esbuild will minify this to 1.toString (removing the zero) which will result in a syntax error in the browser.

### DIFF
--- a/packages/core-js/internals/uid.js
+++ b/packages/core-js/internals/uid.js
@@ -3,7 +3,7 @@ var uncurryThis = require('../internals/function-uncurry-this');
 
 var id = 0;
 var postfix = Math.random();
-var toString = uncurryThis(1.0.toString);
+var toString = uncurryThis(1.1.toString);
 
 module.exports = function (key) {
   return 'Symbol(' + (key === undefined ? '' : key) + ')_' + toString(++id + postfix, 36);


diegovdev:
Proof that it's the same resulting toString method, in the browser console:


image

0f-0b:
esbuild actually minifies `1.0.toString` to `1 .toString` (note the space), so it must be some other tool that misprinted the code.

packages/core-js/internals/uid.js
```diff
@@ -3,7 +3,7 @@ var uncurryThis = require('../internals/function-uncurry-this');
 
 var id = 0;
 var postfix = Math.random();
-var toString = uncurryThis(1.0.toString);
+var toString = uncurryThis(1.1.toString);
```
zloirock, line 6: This is a bug on your minificator side. I already added some workarounds for such kind of minificator bugs **to this line** (for example, `1..toString -> 1.0.toString)`, but I don't think we should encourage these bugs any further. I can accept this PR for compatibility with old versions of this tool, but only in case if this bug will be fixed in this tool itself.

zloirock:
@diego-carvallo I'm curious: Did you miss my comment above? It's not a `core-js` bug, but I'm ready to accept this PR as a workaround in case you at least report it to the source of this bug. Accepting this PR without that is irresponsible since it will cause the same problems in case of such a valid JS expression in other code — and there were already examples of this mentioned above. Is reporting this to the source of the problem so hard to do and easier to throw a tantrum?

diego-carvallo:
Here is the issue reported to ESBuild https://github.com/evanw/esbuild/issues/3975

xximwon (approved):

mcmimik:
Hey, guys,
and all you folks who're still using google

@diegovdev I noticed the "Format FILE" button is activated in your screenshot:

Image

In my case, Chrome (!!) removed that space when I formatted code with that button.
I was using the "Overrides" feature on that file, so that buggy formatter quietly broke the site. Didn't see that coming!

**TL;DR:** It's the known [Chromium DevTools' bug](https://issues.chromium.org/issues/342488683).

Btw:

- ❌ WebStorm 2025.1.1 has this issue too, if you ask it to format the file, it produces `1.toString`.
- ✅ Prettier adds parentheses like `(1).toString` (you can "fix" your IDE with that).
- ✅ Firefox 138 also does it right, formatting without touching the space at all: `1 .toString`.
- ❌ Safari 18.4 fails the format test, dropping the space: `1.toString`.

So we have many places where developers struggle trying to figure out what's wrong with their code. Maybe that 1 byte is really worth it. Given the variety of tools in real-world use—who knows how many transformers disrespect the spaces like this.

---

A few hours later... I found that `1.0.toString` is used in [multiple places](https://github.com/search?q=repo%3Azloirock%2Fcore-js+1.0.+path%3A*.js&type=code) across the repo, not just the file this PR touches.

To make this fix bulletproof, it would make sense to update all modules for consistency. @zloirock @diegovdev What do you think about expanding the PR to cover those cases as well? 100% agree that minifiers like esbuild are the root of this `1 .toString` mess, but I can see why they treat it as still valid code. 😄

And, sorry for that longread, but I was curious how many minifiers and formatters were affected, so I've made some research:

| MINIFIER ⚠️    | `1.0.toString` → | FORMATTER ❌      | 👻 `1 .toString` → |
|----------------|------------------|-------------------|--------------------|
| **terser**     | → `1..toString`  | **prettier**      | → `(1).toString`   |
| **uglify-js**  | → `1..toString`  | **uglify-js**     | → `1..toString`    |
| **swc**        | → `1..toString`  | **biome**         | → `(1).toString`   |
| **esbuild** 👻 | → `1 .toString`  | **deno**          | → `(1).toString`   |
| **bun** 👻     | → `1 .toString`  | **bun** 👏        | → `1 .toString`    |
|                |                  | ***Firefox*** 👏  | → `1 .toString`    |
|                |                  | ***Chromium*** 💀 | → `1.toString`     |
|                |                  | ***Safari*** 💀   | → `1.toString`     |
|                |                  | ***WebStorm*** 💀 | → `1.toString`     |

Interestingly, several commercial tools fail this test, while open-source tools get it right. Just saying.

P.S. Filed bugs with WebStorm & WebKit, see links below.

---

##### References

- [Javascript pretty-printing in Chromium DevTools introduces syntax errors](https://issues.chromium.org/issues/342488683)
- 🆕 [YouTrack: Incorrect code formatting: 1 .toString is turned into invalid 1.toString by "Reformat Code" command (WEB-73189)](https://youtrack.jetbrains.com/issue/WEB-73189/Incorrect-code-formatting-1-.toString-is-turned-into-invalid-1.toString-by-Reformat-Code-command)
- 🆕 [293557 – Web Inspector: "Pretty print" removes required space in 1 .toString, producing invalid syntax](https://bugs.webkit.org/show_bug.cgi?id=293557)

zloirock:
@mcmimik you've convinced me, I'll add a workaround.

PR #1402, opened 2025-03-31
Don't call well-known Symbol methods for RegExp on primitive values

slowcheetah:


zloirock:
Reason: https://github.com/tc39/ecma262/pull/3009

zloirock (approved):

PR #1418, opened 2025-04-19
feat: Add Google Apps Script compatibility runner

chad-loder:
This adds a standalone Google Apps Script runner (`tests/compat/gas-runner.gs`) to test the native feature support of the GAS V8 runtime environment against the checks defined in `tests/compat/tests.js`.

**Motivation:**

Provide a way to easily check the native ES feature compatibility of the Google Apps Script environment, which is relevant for understanding potential polyfill needs when using `core-js` within GAS projects.

**How it Works:**

*   The script is designed to be pasted into the script editor bound to a Google Sheet.
*   It fetches a specific, commit-pinned version of `tests/compat/tests.js` from the `core-js` repository.
*   It validates the fetched file's hash against an expected value.
*   It modifies the fetched code slightly to ensure compatibility with the GAS execution environment.
*   It evaluates the code and executes the test functions defined within it.
*   It logs the results (pass/fail for each native feature check) to a new timestamped sheet within the active Google Sheet, applying conditional formatting for readability.

**Usage:**

Instructions for use are included in the comment header within the `gas-runner.gs` file.

**Limitations:**

*   This script *only* tests the native environment's support for features checked by `tests.js`. It does **not** load or test the `core-js` polyfills themselves.
*   It relies on the GAS environment having network access to `raw.githubusercontent.com`.

This runner can help generate compatibility data specific to the Google Apps Script platform.

### DIFF
--- a/tests/compat/gas-runner.gs
+++ b/tests/compat/gas-runner.gs
@@ -1,0 +1,517 @@
+// @ts-check
+'use strict';
+
+// Declare GAS globals for @ts-check
+/** @type {GoogleAppsScript.Spreadsheet.SpreadsheetApp} */
+const SpreadsheetApp = globalThis.SpreadsheetApp;
+/** @type {GoogleAppsScript.Base.Logger} */
+const Logger = globalThis.Logger;
+/** @type {GoogleAppsScript.Utilities.Utilities} */
+const Utilities = globalThis.Utilities;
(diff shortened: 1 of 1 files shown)


chad-loder:
Results from running the GAS runner script:


Summary: 283 Passed, 129 Failed (as per CSV file). Actual pass/fail counts are reflected in the table above.

PR #1436, opened 2025-05-21
check if toString is writable

olafbuitelaar:
in cases the `Function.prototype` is sealed/frozen, this will throw an error causing any remaining code to abort.

### DIFF
--- a/packages/core-js/internals/make-built-in.js
+++ b/packages/core-js/internals/make-built-in.js
@@ -7,6 +7,7 @@ var DESCRIPTORS = require('../internals/descriptors');
 var CONFIGURABLE_FUNCTION_NAME = require('../internals/function-name').CONFIGURABLE;
 var inspectSource = require('../internals/inspect-source');
 var InternalStateModule = require('../internals/internal-state');
+var getOwnPropertyDescriptor = require('../internals/object-get-own-property-descriptor').f;
 
 var enforceInternalState = InternalStateModule.enforce;
 var getInternalState = InternalStateModule.get;
@@ -16,6 +17,7 @@ var defineProperty = Object.defineProperty;
 var stringSlice = uncurryThis(''.slice);
 var replace = uncurryThis(''.replace);
 var join = uncurryThis([].join);
+var functionToStringDescriptor = getOwnPropertyDescriptor(Function.prototype, 'toString');
 
 var CONFIGURABLE_LENGTH = DESCRIPTORS && !fails(function () {
   return defineProperty(function () { /* empty */ }, 'length', { value: 8 }).length !== 8;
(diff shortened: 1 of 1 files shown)


zloirock:
Why and when could it be frozen?

olafbuitelaar:
If some external code call's; `Object.freeze(Function.prototype);
 Object.seal(Function.prototype); ` prior. Alternativly we could build using core-js-pure, but ideally we can remain building using just core-js, and handle this case specifically.

zloirock:
It's a case of "broken environment". You should load `core-js` before such manipulations with built-ins.

olafbuitelaar:
unfortunate that's not always an option. I would prefer the build bundle with core-js remainings functioning rather the erroring out.

zloirock:
Maybe it's rather an issue for the place where this external code comes from? If I accept your PR in the proposed form, it is tens, if not hundreds, of bytes to millions of other `core-js` users that do not have dependencies, which breaks the environment.

olafbuitelaar:
It's definitely an issue from the external code, unfortunate there is not always a control over this, which is indeed specific for our case. If you have suggestions to minimise the impact on file size. Or otherwise maybe the case could be handle by adding a try catch block arround the assignment?

zloirock:
Can you tell me which project this code comes from?

olafbuitelaar:
we're trying to load our library on a website (https://diep.io), which as first executes this code;
```html
<!doctype html><html><head><script>function freeze_object(x)
      {
        Object.freeze(x);
        Object.seal(x);
        Object.freeze(x.prototype);
        Object.seal(x.prototype);
      }
      freeze_object(WebSocket);
      freeze_object(WebAssembly);
      freeze_object(CanvasRenderingContext2D);
      freeze_object(HTMLCanvasElement);
      freeze_object(Function);
      // freeze_object(Object);
      freeze_object(Path2D);
      freeze_object(ImageData);
      freeze_object(ImageBitmap);
      window.frozenHasFocus = Object.freeze({ hasFocus: document.hasFocus });</script></head></html>
      ```
      
The code of the website is not under our control

zloirock:
`Function#toString` is not the only `core-js` polyfill that will not work here. `freeze_object(Object)` is commented, but `freeze_object(Function)` will break some other `core-js` polyfills. Even if not to say about the early stage proposals, it will break `Function#name` and `Function#[@@hasInstance]` in old engines, `Function#[@@metadata]` cap that's on stage 3, etc..

If you distribute a library, `core-js-pure` is your option. It is precisely intended to prevent such conflicts.

olafbuitelaar:
i agree with that, only `core-js-pure` has some drawbacks, as it increases the overall bundle size, and some Objects aren't native or extandable, like `Promise` as this is implemented as a function rather than a class.

olafbuitelaar:
thank you for your consideration, and general hard work on `core-js`

zloirock:
I thought about this for quite a long time, and I think that this is a direct case for a pure `core-js` version.

PR #1310, opened 2023-11-27
[WIP] V4

zloirock:
## Polyfill packages
- [x] [**Drop ancient engines support**](https://github.com/zloirock/core-js/issues/1301):
  - **The baseline is engines about IE11, ES5 with some additions**:
    - Basic `WeakMap` support (no matter bugs, for internal usage only, in polyfills could be fixed)
    - Basic `Map` and `Set` support (no matter bugs, for internal usage only, in polyfills could be fixed or completely replaced)
    - Basic `%TypedArray%`, `ArrayBuffer` and `DataView` constructors support (no matter bugs, in polyfills could be fixed)
    - A way of setting a prototype - `Object.setPrototypeOf` or `__proto__`
  - Examples of supported engines:
    - IE11
    - Chrome 38
    - Safari 7.1
    - FF15
    - Node 0.11
  - Examples of no longer supported engines:
    - IE10-
    - Android 4.4.3-
    - PhantomJS (~ IIRC Safari 6.0)
    - Opera Presto (12-)
    - Duktape
  - Code changes coming from this:
    - [x] Allow unquoted keyword properties
    - [x] Allow trailing commas in multiline object literals
    - [x] Drop descriptors check
    - [x] Consider strict mode as supported
    - [x] Use native `Array.isArray` without fallbacks
    - [x] Use native `Array.prototype.{ forEach, every, some }` without fallbacks
    - [x] Use native `Date.now` without fallbacks
    - [x] Use native `Date.prototype.toISOString` without fallbacks
    - [x] Use native `Function.prototype.bind` without fallbacks
    - [x] Use native `Object.create` without fallbacks in simple cases
    - [x] Use native `Object.getPrototypeOf` without fallbacks in simple cases
    - [x] Use native `Object.{ defineProperty, defineProperties, getOwnPropertyDescriptor }` without fallbacks in simple cases
    - [x] Use native `Object.keys` without fallbacks in simple cases
    - [x] Use native `Object.getOwnPropertyNames` without fallbacks in simple cases
    - [x] Use native `Object.{ freeze, seal, preventExtensions }` without fallbacks in simple cases
    - [x] Use native `Object.{ isFrozen, isSealed, isExtensible }` without fallbacks in simple cases
    - [x] Consider base `JSON` methods as existent
    - [x] Drop `arguments` fallback from `Object#toString`
    - [x] Drop ES3-related Babel transforms from tests processing
[...]

PR #1442, opened 2025-06-29
Add a minify option to the core js builder.

johnzhou721:
As seen in the main v4 PR, one of the todos is ``Add minify option to @core-js/builder, bundles minified by default`` -- I've decided to add this as a PR against the branch that #1310 is based on.

I'm so sorry that you were/are going through those hard things for maintaining such an important project in the js ecosystem. And sorry for this trivial PR that takes up your time to review.

### DIFF
--- a/packages/core-js-builder/package.json
+++ b/packages/core-js-builder/package.json
@@ -25,9 +25,10 @@
   "main": "index.js",
   "types": "index.d.ts",
   "dependencies": {
-    "core-js": "4.0.0-alpha.0",
+    "webpack": "^5.100.2",
     "@core-js/compat": "4.0.0-alpha.0",
-    "webpack": "^5.100.2"
+    "core-js": "4.0.0-alpha.0",
+    "terser-webpack-plugin": "^5.3.14"
   },
   "engines": {
     "node": "^18.12.0 || >=20.0.0"
--- a/package-lock.json
+++ b/package-lock.json
@@ -50,6 +50,8 @@
     },
     "node_modules/@ampproject/remapping": {
       "version": "2.3.0",
+      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.3.0.tgz",
+      "integrity": "sha512-30iZtAPgz+LTIYoeivqYo853f02jBYSd5uGnGpkFV0M3xOt9aN73erkgYAmZU43x4VfqcnLxW9Kpg3R5LC4YYw==",
       "dev": true,
       "license": "Apache-2.0",
       "dependencies": {
@@ -62,6 +64,8 @@
     },
     "node_modules/@babel/code-frame": {
       "version": "7.27.1",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.27.1.tgz",
+      "integrity": "sha512-cjQ7ZlQ0Mv3b47hABuTevyTuYN4i+loJKGeV9flcCgIK37cCXRh+L1bd3iBHlynerhQ7BhCkn2BPbQUL+rGqFg==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
@@ -75,6 +79,8 @@
     },
     "node_modules/@babel/compat-data": {
       "version": "7.28.0",
+      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.28.0.tgz",
+      "integrity": "sha512-60X7qkglvrap8mn1lh2ebxXdZYtUcpd7gsmy9kLaBJ4i/WdY8PqTSdxyA8qraikqKQK5C1KRBKXqznrVapyNaw==",
       "dev": true,
       "license": "MIT",
       "engines": {
@@ -83,6 +89,8 @@
     },
     "node_modules/@babel/core": {
       "version": "7.28.0",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.28.0.tgz",
+      "integrity": "sha512-UlLAnTPrFdNGoFtbSXwcGFQBtQZJCNjaN6hQNP3UPvuNXT1i82N26KL3dZeIpNalWywr9IuQuncaAfUaS1g6sQ==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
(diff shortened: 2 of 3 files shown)


zloirock (changes requested):
`core-js-builder@3` should work in Node 8 for avoiding breaking changes. The actual `terser` require Node 10. Also, it's even not added to `package.json`.

johnzhou721:
I apologize for this mistake. I will keep looking for more code minifiers, and I'm sorry for adding to your workload.

johnzhou721:
@zloirock I apologize, but I am somewhat confused. This PR is made against the V4 branch which is your other PR, and the v4 PR says
```
 Update @core-js/builder dependencies (Node 8 support no longer block that)
```
and also the webpack version is v5 which, accoring to https://webpack.js.org/migrate/5/, needs node 10.

Do you intend to backport this to V3? If so, I can keep looking for other options, and will an older terser work in that case?

Thank you.

zloirock:
Ah, to v4, sorry... Yes, it will be added to v4, but it's a little more complex. I'm not sure will be used `terser` or something else, even if `terser` will be used - it should use [some options](https://github.com/zloirock/core-js/blob/9d8381d255ecef0cf99fc05be667bf63c8993189/scripts/bundle-package/bundle-package.mjs#L22-L47) and somehow export sourcemap.

Thanks for trying to help, but at this point I have to figure out a few things myself -)

johnzhou721:
> Ah, to v4, sorry... Yes, it will be added to v4, but it's a little more complex. I'm not sure will be used `terser` or something else, even if `terser` will be used - it should use [some options](https://github.com/zloirock/core-js/blob/9d8381d255ecef0cf99fc05be667bf63c8993189/scripts/bundle-package/bundle-package.mjs#L22-L47).
> 
> Thanks for trying to help, but at this point I have to figure out a few things myself -)

You're welcome.  I will research some of the options and see what I can do, but as you said, I might just have to leave it to you.

In addition, what is going to be the future of v4 licensing? I'm totally good with (also it'd be sort of sad to some people) it being commercial etc to force the big companies to pay, but I'm simply curious about the future of the project (FWIW: it's okay for you in my opinion to be undecided on this)

zloirock:
In addition to the availability of a free open source version, the possibility of adding a commercial license is being considered.

johnzhou721:
Great. Thanks for letting me know (when you didn't have to in the first place, mean this nonnegatively.).

johnzhou721:
Also a possiblity is to maybe have the same functionality but make the free version use GPL or some sort of copyleft license s.t. you would force the big companies to open source their code if they don't pay whatever price you want them / need them to... given that a lot of projects are only offered under the GPL, that seems fair probably to a lot of people.

That said, there's the probability of people just stealing the GPL code and not getting a license...

Closing the loop on this.  Sorry for the off-topic (to this PR) discussion.

johnzhou721:
Nevermind... we need to map to the original source files...

johnzhou721:
FYI: feel free to reuse any code from here.

johnzhou721:
@zloirock FWIW this is very hard. especially with the replacing part and the need to generate a source map. if you are able to use the output of webpack as-is without doing the extra things it should be easy.

Sorry for being unable to help here.

johnzhou721:
@zloirock Hmm... not familiar with the structure of sourcemaps for JS... but someone who *is* might be able to pull this off, but not sure how hard it's going to be.

Webpack *can* generate minified + sourcemap; see 31239cc3b.

However, that is non-working because we gotta have this huge banner summary on top and replacing function __webpack_require__ or something like that with __webpack_require__ = function or the other way around, and we're also wrapping everything in a huge anonymous function that has undefined as argument.

Question:
- Is the first change I mentioned purely cosmetic, or are we assinging to __webpack_require__ later?
- The second change is probably to prevent overwriting undefined. If we add it on a seperate line, we can just shift the line numbers in the sourcemap (not sure how to do).

If I get time (I have other massive PRs underway), I will try to see what I can do about understanding how those source .map files work.

johnzhou721:
Ok... looks like minimum ECMAScript version is 5 in v4, which means this just got a whole lot easier, because ES5 does not allow you to overwrite undefined if I remember correctly, and webpack with terser already adds a !function(){}() around stuff, so we just need the preamble. Not sure if the function __webpack_require__() vs var __webpack_require__() = function thing matters, because we're not overwriting the function nor calling it before the declaration, so I omitted that change.

johnzhou721:
Question: Somehow a bunch of stuff is getting removed from package_lock.json, how am I supposed to handle that? What command I run to update that JSON but not have it produce this huge diff?

zloirock:
At least, you had removed `webpack` from dependencies.

johnzhou721:
Annnnnd... Green CI!!!

@zloirock A couple of questions:
- Is it acceptable to not have the source map if not minified?
- How would I add tests to verify that all usages of the builder work correctly?  Is using a bundle done with different settings to test in various workflows enough?
- Do you object to not replacing with `var __webpack_require__() = function` when minified? ``webpack-plugin-replace`` might work here, but I'd need to dig into it.

Thank you!

johnzhou721:
Conflicts should be resolved now.

packages/core-js-builder/package.json
```diff
@@ -25,8 +25,10 @@
   "main": "index.js",
   "types": "index.d.ts",
   "dependencies": {
-    "core-js": "4.0.0-alpha.0",
     "@core-js/compat": "4.0.0-alpha.0",
+    "core-js": "4.0.0-alpha.0",
+    "konan": "^2.1.1",
```
zloirock: Why is this here?

johnzhou721:
```suggestion
```
I don't know... probably some sort of merge resolution.  Thanks for the catch.

johnzhou721:
Forgot to ask for a rereview… will do so after I get back home and rebase.

johnzhou721:
@zloirock I tried using some random 3rd party plugins to achieve the replacement and the wrapping, but my attempts has failed... so I need more info on what the ``= function`` replacement is supposed to do, and whether the ``!function(undefined)`` is a historical thing... ES5 doesn't allow you to replace undefined anymore, so the latter should not be a concern. Thanks!

zloirock:
`!function(undefined)` allows minification `undefined` to 1 byte, similarly like `= function` for `__webpack_require__`.

johnzhou721:
@zloirock  Wow... can't believe no one's still willing to pay you to maintain core-js given your dedication to speeding up every second for users of this and also your cleverness in doing it...

I think we can do
- webpack (without optimization)
- magic-string (to replace the stuff)
- terser

Each of these steps will produce a source-map, and we need to combine it. https://stackoverflow.com/questions/21709341/combine-source-maps-of-two-compilation-steps might be a good reference... sorcery makes it easy, but the README says it's a work-in-progress... @ampproject/remapping
 is used by Babel, but I need to study how it works.... and guess what, I've ran out of time.  Sorry again.

PR #1446, opened 2025-07-03
Correct path to get-built-in.js

johnzhou721:
Documentation improvement.

@zloirock Question: In v4 it is suggested to use ``// dependency`` comments to specify deps, is that still true in v3?

Sorry for the slowness. I'm still trying to figure out the architechture of this project.


johnzhou721:
Oh, nevermind, use ``require('../../modules/esnext.regexp.escape');`` entires to specify such things.

I'm stupid...

zloirock:
> In v4 it is suggested to use // dependency comments to specify deps, is that still true in v3?

No, it's (parser such comments and entries / compat data generator from this data) added in v4.

zloirock:
Thanks.

zloirock (approved):

PR #1448, opened 2025-07-03
Temporal Impl

johnzhou721:


johnzhou721:
FYI: I've just realized that the work-in-progress temporal shim posted in the issue is very far from complete, and that the latest Temporal polyfill is extremely massive and also relies on es-abstract, which has lots of methods that will take forever for me to figure out the corresponding helper functions in core-js. Closing on that basis, as I realized that even if I just translate the polyfill to ES3, it'd take a long amount of time that I don't and won't be having in a lot of years.

Sorry for the noise.

PR #1462, opened 2025-09-18
[NFC] Remove unused postinstall for core-js-pure

pnappa:
There is no postinstall.js file in this directory, so the postinstall is always a no-op.

---

Reasoning: `pnpm` only allows whitelisted postinstall scripts to run by default, and emits a warning message. Due to the recent supply chain attack exploiting postinstall scripts, it's a good idea to only allow safe postinstall scripts, and the less noise from unused ones, the better.

Thus, I've removed the unused postinstall script for core-js-pure, as it doesn't do anything (it just try/catches the require of a non-existent file). I have not done the same for the other packages, as they're clearly using a postinstall.js script.

### DIFF
--- a/packages/core-js-pure/package.json
+++ b/packages/core-js-pure/package.json
@@ -72,8 +72,5 @@
   "sideEffects": [
     "./modules/*.js"
   ],
-  "main": "index.js",
-  "scripts": {
-    "postinstall": "node -e \"try{require('./postinstall')}catch(e){}\""
-  }
+  "main": "index.js"
 }


zloirock:
Thanks for the PR, however, it's copied from another place. You could find it in the package directly.

zloirock:
Note that it will be removed and changed to another approach at `core-js@4` release.

pnappa:
Oh, thanks! I only saw that it was copied to 'core-js-bundle' in a script, I didn't see any similar code for core-js-pure.

But, I can definitely see it in https://www.npmjs.com/package/core-js-pure?activeTab=code - so you're definitely 100% right.

Apologies for the disruption!

Issue #1468, opened 2025-10-09
Remove `Function.prototype.toString` global monkey patch or make it opt-in


lionel-rowe:
As a library consumer, I don't want any library I import to monkey-patch globals unless I explicitly opt in to the behavior. For example, if I import a polyfill for `Iterator.zip`, I don't mind so much if an extant `Iterator.zip` static method gets overridden, but I definitely don't want `Function.prototype.toString` to also be clobbered.

In addition, I don't want core-js functions to serialize as if they're native; this makes it harder to debug whether a given polyfill is active. For example, if I log `Iterator.zip.toString()`, I only want it to show as `function zip() { [native code] }` if it's actually native (i.e. _not_ polyfilled).

Currently, importing from many (most? all?) core-js entrypoints globally overwrites `Function.prototype.toString` via their dependency on `internals/make-built-in.js`:

https://github.com/zloirock/core-js/blob/ce89bd7c00eefb90b5c8cea12ba82d1e121cb4d6/packages/core-js/internals/make-built-in.js#L51-L55

I'm not sure what the original rationale was for trying to trick lodash into thinking core-js's polyfills are native, but in any case, lodash now intentionally counteracts this behavior by making `_.isNative` throw if it detects the presence of core-js:

https://github.com/lodash/lodash/blob/ddfd9b11a0126db2302cb70ec9973b66baec0975/lodash.js#L11899-L11930

```js
import 'core-js/modules/esnext.iterator.zip.js'
import _ from 'lodash'
_.isNative(Array.prototype.push)
// Uncaught Error: Unsupported core-js use. Try https://npms.io/search?q=ponyfill.
//     at lodash.isNative (file:///path/to/lodash/4.17.21/lodash.js:11995:15)
//     at <anonymous>:3:3
```

As a result, I don't see how there can be any possible rationale for keeping this. Any original benefit alongside lodash usage is now worse-than-useless, as it prevents lodash from being able to identify _any_ function as native, even actual native functions.

zloirock:
[...]

PR #1471, opened 2025-10-09
Remove global `Function#toString` monkey patch

lionel-rowe:
Fixes https://github.com/zloirock/core-js/issues/1468.

I'd already written most of this PR before that issue got closed — I've left a comment in the issue explaining why I don't think it should be closed.

An alternative (partial) solution would be selectively overwriting `toString` of only core-js functions, which would at least eliminate the need for the global monkey patch, even though it still wouldn't solve the problem of degraded debugging experience. Let me know if that would be a preferable solution.

---

Instead of testing for `/native code/` in the monkey-patched serialized source, we now use the new utilities under `function-kind.js` for internal checking of whether a function comes from core-js, the platform, or elsewhere. In order of preference, core-js functions are stored in a weakmap, added as a unique symbol property, or added as a string property, depending on availability of `WeakMap` and `Symbol`. The new check for true native functions should now be more reliable too, e.g. it won't return false positives for third-party functions like `() => "{ [native code] }"`, though it can still return false positives if third-party code overrides the `toString` method of its functions.

Registration of core-js functions and detection of true native functions is handled through `shared-store` to avoid conflict with multiple versions (even if one version is older and regardless of loading order).

### DIFF
--- a/packages/core-js/internals/function-provenance.js
+++ b/packages/core-js/internals/function-provenance.js
@@ -1,0 +1,43 @@
+'use strict';
+var inspectSource = require('./inspect-source');
+var store = require('./shared-store');
+var isCallable = require('./is-callable');
+
+if (!isCallable(store.getFunctionProvenance)) {
+  // https://tc39.es/ecma262/#prod-NativeFunction
+  var NATIVE_CODE_RE = /\)\s*\{\s*\[\s*native\s+code\s*\]\s*\}\s*$/;
+  // eslint-disable-next-line es/no-weak-set -- safe
+  var CORE_JS_REGISTRY = typeof WeakSet === 'function' ? new WeakSet() : null;
+  // eslint-disable-next-line es/no-symbol -- safe
+  var CORE_JS_BRAND = typeof Symbol === 'function' ? Symbol('CORE_JS_BRAND') : '__CORE_JS_BRAND__';
+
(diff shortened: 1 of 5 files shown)


zloirock:
Why removing of `Function.prototype.toString` patch is a bad idea, I had explained in the issue.

As you could see, for debugging, you could use `globalThis['__core-js_shared__'].inspectSource`.

lionel-rowe:
> @zloirock As you could see, for debugging, you could use globalThis['__core-js_shared__'].inspectSource.

I have two separate concerns. The more important one is that I don't want any library monkey patching global methods, ever (unless I explitly opt into it).

> @lionel-rowe An alternative (partial) solution would be selectively overwriting `toString` of only core-js functions, which would at least eliminate the need for the global monkey patch

Would this alternative solution be acceptable? I can open a PR with this if that would be preferable. The serialization of core-js functions would remain identical to what it is now, just without any monkey patching.

zloirock:
> Would this alternative solution be acceptable? I can open a PR with this if that would be preferable. The serialization of core-js functions would remain identical to what it is now, just without any monkey patching.

The thing is that problematic libraries use something like `/native code/.test(Function.prototype.toString.call(it))`, not just `/native code/.test(it)`, for feature detection, so I don't see alternatives to patching.

Issue #1497, opened 2026-01-13
Use CreateDataProperty instead of assignment in array iteration methods


arcde40:
Hi, I'm a student studying ECMAScript and Test262. 

While testing, I noticed that some `Array.prototype` methods are failing specific Test262 cases. For example, `Array.prototype.map` fails about 5%(22/429) of the test cases.

### Problem
According to a [Spec](https://tc39.es/ecma262/#sec-array.prototype.map), `Array.prototype.map` should use `CreateDataPropertyOrThrow()` to define values. 
However, the current implementation is using simple assignment(using setter). 

```javascript
// internals/array-iteration.js
//...
for (;length > index; index++) if (NO_HOLES || index in self) {
  value = self[index]; 
  result = boundFunction(value, index, O);
  if (TYPE) {
	if (IS_MAP) target[index] = result; // Should be CreateDataPropertyOrThrow()
	else if (result) switch (TYPE) {
		// ...
```

When I replaced the assignment with `createProperty()` from `internals/create_property.js`, it passed most of the previously failing tests.

### Performace bench

Since `CreateDataProperty` is a heavy function compared to simple assignment, I've ran some simple benchmarks to compare the performance:

Environment: Node v24.11.0 / MacBook Pro (M1 Pro)

| Array Length | Method        | Mean Time | Difference    |
| ------------ | ------------- | --------- | --------- |
| 10 |Current| 174.42 ms | - |
|              |`createProperty()`  | 173.83 ms | Very small |
| 100 |Current| 177.61 ms | - |
|              |`createProperty()`| 177.16 ms | Very small |
| 1,000,000 |Current| 390.81 ms | - |
|              |`createProperty()`| 613.39 ms | ~1.57x slower |

Although there's a performance drop for extreme cases (1M length), the overhead is negligible for common array sizes.

### Conclusion
This issue seems to affect other methods as well, such as `filter`, `flat` and `flatMap`.

Is the current implementation intentional(for performance)? Considering the benchmark results, I believe prioritizing spec compliance is the right direction since there's no significant performance issue in real-world scenarios.

And if you agree, could I submit a PR to resolve this issue? (This would be my first contribution to open source, so I would appreciate your guidance!)

Thank you for your hard work on this project!

zloirock:
Yes, you are right, this is because of the performance reasons. In ancient engines, `defineProperty` was much slower - however, I'm not sure about your performance bench even for modern.

[...]

PR #1498, opened 2026-01-14
Fix some Array iteration methods to use CreateDataProperty

arcde40:
## Summary
This PR updates `Array.prototype.filter` (and `filterReject`), `flat`, and `flatMap` to use `CreateDataProperty` instead of `push` or assignment.
This ensures the implementation follows the ECMAScript specification, and resolves Test262 failures.
Related to #1497

### 1. `filter` and `filterReject`
- Change: Replaced `push` with `createProperty` (CreateDataProperty).
- Reason: The previous implementation using `push` could trigger setters on `Array.prototype`, causing Test262 failures.

### 2. `flat` and `flatMap`
- Change: Updated the `flattenIntoArray` to use `createProperty`.
- Change: Removed the explicit `length` assignment in `flat` and `flatMap`.
- Reason: According to the spec, `flat` and `flatMap` do not explicitly set the length of the result array.

##  Note on `Array.prototype.map`
`Array.prototype.map` also uses assignment instead of `CreateDataProperty`. 
I considered including a fix for `map` as well (since it was the main purpose of the #1497), but decided to exclude it from this PR for the following reasons:

1.  The `map` logic in `array-iteration.js` is shared with TypedArrays. TypedArrays require `Set` for `map`, whereas standard Arrays require `CreateDataProperty`. ([%TypedArray%.prototype.map](https://tc39.es/ecma262/#sec-%typedarray%.prototype.map) vs [Array.prototype.map](https://tc39.es/ecma262/#sec-array.prototype.map))
2. Without strictly checking whether the target is a TypedArray or an Array, it is impossible to comply with both specs simultaneously.
3. Strict type checking would require extra code (e.g., `ArrayBuffer.isView`), resulting in increased bundle size and overhead.

Since this seems to be an architectural trade-off, I would appreciate your guidance on the preferred approach regarding this trade-off.

---


zloirock:
> Reason: According to the spec, flat and flatMap do not explicitly set the length of the result array.

It seems it was removed in the latest spec revisions.

zloirock:
Thanks.

PR #1505, opened 2026-01-29
Document atob and btoa

johnzhou721:
As title says, this adds docs for atob and btoa functions


zloirock:
It's here: https://core-js.io/docs/features/web-standards/base64-utility-methods

Issue #638, opened 2019-09-04
Build next generation polyfill service using `core-js` polyfills


neenhouse:
# Overview / RFC

(Work in progress)

Looking forward, we'd like to build a next generation polyfill service that uses corejs polyfills as it's core to capture next gen performance.  As stated from @zloirock a look to the future:

> A service like this one integrated with a good polyfills source like core-js, which only loads the needed polyfills by statically analyzing the source like Babel's useBuiltIns: usage option does could cause a revolution in the way we think about polyfills.

We've tested and built this service and wish to open source it back to community with the help and input of @zloirock.  We should be able to share the details of the implementation in a RFC here and are actively working on open sourcing the internal components we have already built.  We intend to publish a demo of this service to `corejs.io` as a proof of concept and were hoping to co-maintain this package as an extension of corejs.io. 

Lots of details to document here and discussion to happen in the next few days.

zloirock:
Sorry for the delayed reply, I have some problems with time. Sounds great. Waiting for updates...

neenhouse:
Also apologize for the delay on our end.  We should have code published soon.

slavafomin:
@neenhouse this looks extremely interesting, we would be very glad to see such service implemented. Do you have more progress? Is there a way for us to help?

neenhouse:
@slavafomin Yes!  We fully intend to do all phases in the open.   Since releasing our code has taken longer than anticipated, I would be open to meeting to discuss our next steps.  I'm wondering if @zloirock would be open to meeting up with us as well?

zloirock:
@neenhouse sorry for the delay, I had (and still have) some serious problems, but let's try to find a convenient time for both of us. Could you mail me a convenient time for you?

Awen-hub:
> # Overview
> Looking forward, we'd like to build a next generation polyfill service that uses corejs polyfills as it's core to capture next gen performance. As stated from @zloirock a look to the future:
> 
> > A service like this one integrated with a good polyfills source like core-js, which only loads the needed polyfills by statically analyzing the source like Babel's useBuiltIns: usage option does could cause a revolution in the way we think about polyfills.
> 
> We've tested and built this service and wish to open source it back to community with the help and input of @zloirock. We should be able to share the details of the implementation in a RFC here and are actively working on open sourcing the internal components we have already built. We intend to publish a demo of this service to `corejs.io` as a proof of concept and we're hoping to co-maintain this package as an extension of corejs.io.
> 
> # Motivation
[...]

PR #1506, opened 2026-02-04
Rework `@core-js/builder`

### LINKED ISSUE #1272: Replace `webpack` in `core-js-builder`
core-js-builder uses webpack as its dependency. Because webpack has many dependencies, introducing core-js-builder in a project will indirectly introduce many dependencies of webpack, which increases project management costs.

Can we use rollup to replace the webpack in core-js-builder? Thank you so much.

zloirock:
- Replace Webpack with Rolldown
- Since we need support of pre-ES6 engines and Rolldown contain ES6 syntax in helpers / wrappers, transpile the output with SWC
- Add `minify` option with SWC (enabled by default, will think about which option to leave as the default.)
- Use SWC only in cases of minification or target engines without support for the required syntax
- Depends on cases, that makes the builder 5x-20x faster
- This speedup is a required step for the implementation of service, #638
- Added some mechanics (`__external` in `@core-js/compat`) that could be used in the future for targets-based syntax transpiling of the polyfill
- Since Rolldown is distributed as ESM, move the package to ESM
- Clarify supported Node version
- Since `utils.styleText` was added in those Node versions, use it instead of direct codes for color-coded `console` output
- Since v4 drops support of some ancient engines, some minification tricks (like splitting bundle to lines lower than 32kb) no longer required

Fixes #1272

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1082,18 +1082,18 @@
       }
     },
     "node_modules/@oxc-project/types": {
-      "version": "0.111.0",
-      "resolved": "https://registry.npmjs.org/@oxc-project/types/-/types-0.111.0.tgz",
-      "integrity": "sha512-bh54LJMafgRGl2cPQ/QM+tI5rWaShm/wK9KywEj/w36MhiPKXYM67H2y3q+9pr4YO7ufwg2AKdBAZkhHBD8ClA==",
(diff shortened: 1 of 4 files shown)


slavbar:
thank you so much

PR #1510, opened 2026-02-08
feat: Improve performance of Uint8Array Hex functions

johnzhou721:
Refs #1503 for the source of inspiration to do this.  This PR significantly improves the performance of Uint8array hex functions.

<details>

<summary> Benchmarks / Console logs on IE11 for from hex (including final performance of from-hex functions in addition to all the moments used for the decisions of what do put) </summary>

```
// Define the functions for the two methods: parseInt and manual hex-to-decimal with bitwise operations

// parseInt method
function parseIntMethod(hex) {
  return parseInt(hex, 16);
}

// hexCharToDecimal method with bitwise operations
function hexCharToDecimal(hex) {
  var firstDigitValue = (hex.charCodeAt(0) >= 48 && hex.charCodeAt(0) <= 57)  // '0' to '9'
    ? hex.charCodeAt(0) - 48 
    : (hex.charCodeAt(0) >= 97 && hex.charCodeAt(0) <= 102)  // 'a' to 'f'
    ? hex.charCodeAt(0) - 87 
    : hex.charCodeAt(0) - 55;  // 'A' to 'F'
    
  var secondDigitValue = (hex.charCodeAt(1) >= 48 && hex.charCodeAt(1) <= 57)  // '0' to '9'
    ? hex.charCodeAt(1) - 48 
    : (hex.charCodeAt(1) >= 97 && hex.charCodeAt(1) <= 102)  // 'a' to 'f'
    ? hex.charCodeAt(1) - 87 
    : hex.charCodeAt(1) - 55;  // 'A' to 'F'
  
  return (firstDigitValue << 4) | secondDigitValue;  // Bitwise operation to combine digits
}

// Function to run the benchmark
function runBenchmark() {
  var testCases = ["aa", "Af", "Bf", "0f", "bA"];
  var iterations = 10000;
  
  // Test with parseInt
  console.time("parseInt");
  for (var i = 0; i < iterations; i++) {
    for (var j = 0; j < testCases.length; j++) {
      parseIntMethod(testCases[j]);
    }
  }
  console.timeEnd("parseInt");

  // Test with hexCharToDecimal (Bitwise)
  console.time("hexCharToDecimal");
  for (var i = 0; i < iterations; i++) {
    for (var j = 0; j < testCases.length; j++) {
      hexCharToDecimal(testCases[j]);
    }
  }
  console.timeEnd("hexCharToDecimal");
}

// Run the benchmark
runBenchmark();
undefined
parseInt: 27.6ms
[...]

PR #1514, opened 2026-02-20
Fix for #810

johnzhou721:
Refs #810 by calling custom implementation on sticky flag.  This may only fix part of the issue.


johnzhou721:
@zloirock Please take a look at this when you've got a chance (along with #1513, #1510).  Thanks!

packages/core-js/modules/es.string.replace.js
```diff
@@ -76,10 +76,12 @@ fixRegExpWellKnownSymbolLogic('replace', function (_, nativeReplace, maybeCallNa
       var rx = anObject(this);
       var S = toString(string);
 
+      var flags = toString(getRegExpFlags(rx));
```
zloirock, line 81: Moving it up without the rest causes incorrect arguments validation / conversion order.

Issue #1456, opened 2025-08-30
Some new polyfills


unhighghlow:
Unfortunately I'm not an experienced javascript developer, nor can I tell if the code I provided is ES3, so I'm submitting these in the form of an issue instead of a PR. With these (passed through babel with [this script](https://codeberg.org/highghlow/nowebsite/src/commit/c1aaf309961c61c718b4a0d4589579dc677345ce/preprocess/preprocess.py#L112-L164)) I was able to run a fairly complex web app ([the nowebsite example](https://codeberg.org/highghlow/nowebsite/src/branch/main/example)) on IE>=9, Chrome>=10 and Firefox>=4 (see: [my tests](https://codeberg.org/highghlow/nowebsite/src/commit/f19c9222429fc892675a6b7324bfcbd954bab58b/README.md#supported-browsers))

A few notes:
1. Most of these weren't written by me. I provide a source in that case
2. There're versions of these polyfills for IE8 (modified by me), but they only barely work on IE8, don't make the app functional (just allow it to render a single page without crashing), and require a [nasty patch to core-js](https://codeberg.org/highghlow/nowebsite/src/commit/ea13d21b6faadd549e87bf1a694d216bab18da8a/preprocess/preprocess.py#L218-L227) (though that is only for the code generated by the bundler, so maybe it's fine)
3. **The most important one of these is requestIdleCallback, which has been requested in #846**

`Element.classList` [(caniuse)](https://caniuse.com/classlist) -- [source](https://github.com/eligrey/classList.js/blob/ecb3305730a9b62f31cfdc71190d2f312233ed32/classList.js): [Unlicense](https://github.com/eligrey/classList.js/blob/ecb3305730a9b62f31cfdc71190d2f312233ed32/LICENSE.md), [ie9+ version](https://codeberg.org/highghlow/nowebsite/src/commit/f19c9222429fc892675a6b7324bfcbd954bab58b/noact/lib/classlist.js), [ie8 version](https://codeberg.org/highghlow/nowebsite/src/commit/ea13d21b6faadd549e87bf1a694d216bab18da8a/noact/lib/classlist.js), [diff](https://codeberg.org/highghlow/nowebsite/commit/ea13d21b6faadd549e87bf1a694d216bab18da8a#diff-744c652d27f045aa3cbc2d65d8b7045f3a9e95fa)

[...]

PR #1524, opened 2026-03-18
Element#remove Polyfill

johnzhou721:
@zloirock Please take a look when you have time — thanks.

There's some other things with ChildNode mixin that have methods like remove, however I think they should be in separate modules; I may submit a PR with these later

Refs #1456.

TODO:
- [x] Test invalid ``this``
- [x] Document


johnzhou721:
@zloirock Can you rerun Karma?  It seems intermittent.

johnzhou721:
@zloirock Take a look at this whenever you're ready -- sorry for the premature request for review.

I've been holding off on implementing the same method for DocumentType etc. because I don't have so much time to create files, so only the Element polyfill for now.

packages/core-js/modules/web.element.remove.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+// TODO: Remove from `core-js@4`
```
zloirock: ?

packages/core-js/modules/web.element.remove.js
```diff
@@ -0,0 +1,33 @@
...
+var wellKnownSymbol = require('../internals/well-known-symbol');
+var create = require('../internals/object-create');
+var defineProperty = require('../internals/object-define-property').f;
+
+if (globalThis.Element) {
+  var BASIC = !!Element.prototype.remove;
+  var removeChild;
+  if (Element.prototype.removeChild) removeChild = uncurryThis(Element.prototype.removeChild);
+
+  $({ target: 'Element', proto: true, forced: !BASIC }, {
```
zloirock, line 17: Please, add a link to the spec.

zloirock (commented):
I'm not sure that this is a good idea to add it right now. Now `core-js` contains only ES and cross-platform features; DOM elements are not cross-platform. There are many things to consider. Will see.

johnzhou721:
@zloirock If yo consider this to be good for core-js, then please consider merging it.  If not, feel free to close.

johnzhou721:
merged with main; let me know if anything else is needed on this PR.

PR #1528, opened 2026-03-20
feat(types):

nmccready:
Add comprehensive .d.ts type augmentations for the most commonly used core-js polyfills covering es.array.*, es.object.*, es.promise.*, and es.string.* module groups. Includes type test file verifying compilation.

### DIFF
--- a/packages/core-js/index.d.ts
+++ b/packages/core-js/index.d.ts
@@ -1,0 +1,381 @@
+// Type definitions for core-js 3.49
+// Project: https://github.com/zloirock/core-js
+// Definitions for the global polyfills provided by core-js
+//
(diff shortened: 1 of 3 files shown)


zloirock:
Comprehensive? This PR does not contain anything that's missed in standard TS types.

`core-js` contains about half a thousand polyfills, many thousands of entry points with their own signatures, a pure version that should not pollute the global namespace, including types. Storing these types, which overlap standard TS types, next to files is extremely unstable, as they will break with almost every minor TS release.

Adding types is a fundamental task.

[It's already added in the V4 branch](https://github.com/zloirock/core-js/tree/v4/packages/core-js-types), a new alpha with types will be published in a few days.

PR #1533, opened 2026-04-02
Cache URL#toString

johnzhou721:
Caches this method so if someone replaces URL#toString after they import core-js, things will not break.


packages/core-js/modules/web.url.to-json.js
```diff
@@ -1,11 +1,12 @@
 'use strict';
 var $ = require('../internals/export');
 var call = require('../internals/function-call');
+var toString = URL.prototype.toString;
```
zloirock: It will not work in the pure version and old engines.

johnzhou721: Hmm... so is this problem perhaps repairable?  Or is it a fundamental limitation of old engines?

Can you be more specific about why this won't work in pure?

zloirock: `URL` is missed in old / ES-only engines. Global `URL` is missed after polyfilling in the pure version.

zloirock: You could use, for example, `get-built-in-prototype-method`.

johnzhou721:
@zloirock This is ready for another look from you — Thanks!!!

zloirock (approved):

Issue #1534, opened 2026-04-07
ActiveXObject fallback in core-js


ziacik-resco:
Hi,

I noticed `core-js` is pulling `ActiveXObject("htmlfile")` into my bundle.

Would it be possible to remove this fallback completely, or provide some way to opt out of including this kind of legacy IE-specific code?

IE is no longer used in practice, and having this code in modern bundles creates issues during security audits.

zloirock:
It's a part of `Object.create` fallback. You could take a look at what depends on it and exclude those modules. However, some modules for modern features depend on it. In core-js@4, `Object.create` fallback will be completely removed and for untangling, some similar cases will be moved to implicit dependencies that will be able to be optimized by transpilers.

Issue #1395, opened 2025-03-03
object-create过于依赖html


rabit2022:
iframe.style.display = 'none';
        html.appendChild(iframe);
        iframeDocument = iframe.contentWindow.document;
iframeDocument.open();
        iframeDocument.write(scriptTag('document.F=Object'));
        iframeDocument.close();

这些代码都会报错，es3非浏览器环境下，没有处理iframe={}的情况，html为空的情况，而且为什么创建对象的方法，会与html产生依赖，是否可以优化

johnzhou721:
Translation:

------

# object-create will error in non-browser environments
```
iframe.style.display = 'none';
html.appendChild(iframe);
iframeDocument = iframe.contentWindow.document;
iframeDocument.open();
iframeDocument.write(scriptTag('document.F=Object'));
iframeDocument.close();
```
This branch of code will error, under a non-browser es3 environment, there's no processing of the case where iframe={}, and html is empty.
```
NullProtoObject = typeof document != 'undefined' ?
document.domain && activeXDocument ?
NullProtoObjectViaActiveX(activeXDocument) // old IE
:
NullProtoObjectViaIFrame() :
NullProtoObjectViaActiveX(activeXDocument); // WSH
```
No more situations considered

gnh1201:
It is likely that an attempt was made to use the "htmlfile" object in a non-browser environment.

In my case, I discovered that on one of the four computers I regularly use to check the results of my development, it was not possible to create an "htmlfile" object.

I suspect that software interacting with internal systems has damaged the operating system’s registration related to the htmlfile object. (The most likely culprit is security software.)

On PCs where the htmlfile object has been damaged, the same issue occurs, and since this is not a common case, I am still looking into how to fix it. Attempting to modify the known related registry entries did not resolve the problem.

In fact, to speak more fundamentally, in a non-browser environment it is unnecessary to import an iframe object and define Object.create() in the process. Even if the relevant operating system component is damaged, there should be no error.

However, it seems that the IE-related code snippets inside core-js were designed with “web browsers” in mind, and thus did not account for separate JavaScript runtimes that have no need to use a DOM with IE-like specifications.

gnh1201:
Previously, I submitted a PR regarding an issue related to that code snippet (#966, #970).

If you could provide a more detailed description of the problem you are currently experiencing, I will consider creating and submitting an additional PR.

zloirock:
[...]

PR #1536, opened 2026-04-09
Crash fix in WSH runtime

gnh1201:
The existing code forces the invocation of the `htmlfile` module even in the WSH environment, which can cause crashes on machines where calling the `htmlfile` module is restricted (e.g., due to unknown changes in IE-related settings).

In the case of WSH (or`non-browser ES3 environment`), its usage differs significantly from that of IE (as it is a standalone runtime rather than a web browser), so there is no real need to invoke the `htmlfile` module. The issue can be resolved by not calling `htmlfile` at all in the WSH environment.

The related issue is as follows:
* #1395
* #1534
* https://github.com/gnh1201/welsonjs/issues/308

### DIFF
--- a/packages/core-js/internals/object-create.js
+++ b/packages/core-js/internals/object-create.js
@@ -22,12 +22,36 @@ var scriptTag = function (content) {
 
 // Create object with fake `null` prototype: use ActiveX Object with cleared prototype
 var NullProtoObjectViaActiveX = function (activeXDocument) {
-  activeXDocument.write(scriptTag(''));
-  activeXDocument.close();
-  var temp = activeXDocument.parentWindow.Object;
+  var temp;
+  try {
+    activeXDocument.write(scriptTag(''));
+    activeXDocument.close();
+    temp = activeXDocument.parentWindow.Object;
+  } catch (error) {
+    temp = NullProtoObjectViaSc32bit();
+  }
   // eslint-disable-next-line no-useless-assignment -- avoid memory leak
   activeXDocument = null;
   return temp;
+};
+
+// Attempt to create a null-prototype-capable Object constructor
+// using the 32-bit-only MSScriptControl (ActiveX).
+// This will fail on 64-bit runtimes or when ScriptControl is unavailable.
+var sc32bit;
+
+var NullProtoObjectViaSc32bit = function() {
+  try {
+    sc32bit = new ActiveXObject("MSScriptControl.ScriptControl");
+    sc32bit.Language = "JScript";
+
+    // Return the Object constructor from the ScriptControl context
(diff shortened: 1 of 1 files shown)


gnh1201:
#970 #966

packages/core-js/internals/object-create.js
```diff
@@ -61,7 +61,7 @@ var NullProtoObject = function () {
     ? document.domain && activeXDocument
       ? NullProtoObjectViaActiveX(activeXDocument) // old IE
       : NullProtoObjectViaIFrame()
-    : NullProtoObjectViaActiveX(activeXDocument); // WSH
+    : function () { return {}; }; // WSH or standalone ES3 runtime
```
zloirock: Returning an object with a prototype is not a fix.

gnh1201: However, if we don’t do this, we can’t avoid application crashes. Is there a better approach?

gnh1201:
For now, I’ve only addressed the app crash issue. I’ll update the code shortly with improvements that more closely align with the intended purpose of `Object.create`.

gnh1201:
On 32-bit systems only (64-bit is not supported), there is a way to obtain a fresh `"Object"` constructor. I implemented this in my changes as a function called `NullProtoObjectViaSc32bit`.

If we cannot give up null-prototype objects, using the `NullProtoObjectViaSc32bit` approach could be a viable option.

gnh1201:
I have added a fallback for 64-bit systems.

packages/core-js/internals/object-create.js
```diff
@@ -22,14 +22,55 @@ var scriptTag = function (content) {
...
+    sc64bit = new ActiveXObject('ScriptControl');
+    sc64bit.Language = 'JScript';
+    return sc64bit.Eval('Object');
+  } catch (error) {
+    // If all attempts to retrieve a new object from an external instance fail,
+    // the existing object is returned.
+    // This prevents errors during the key deletion process using enumBugKeys afterward.
+    return function () {
+      return {};
+    };
```
johnzhou721: I think a better alternative here is to just raise an error; it already raises an error in the status quo, but a better error message ('cannot create object with null prototype in current environment') may be good

gnh1201: That sounds good.

johnzhou721:
@zloirock Since people have asked about 3 tickets about this, can we try to accept this PR?  ActiveX ScriptControl in WSH 32-bit ships out-of-the-box, 64-bit can be installed by a third-party, and this also catches any exceptions and adds a better error message.

@gnh1201 This looks pretty good.  Excited to see people dig *this* deep to get foriegn objects in JavaScript

PR #1541, opened 2026-05-31
feat: Add baseline test infrastructure for Temporal proposal

Raymondo97:
This PR introduces the initial testing infrastructure and baseline specification tests for the upcoming **Temporal** proposal.

To ensure the test suite is manageable and easy to review, this initial PR focuses strictly on setting up the shared harness infrastructure, upgrading the build scripts to support future test suites, and providing a single flat baseline test for the Temporal namespace. 

### Key Changes

* **Build & Tooling Upgrades:**
  * Updated `.gitignore` to prevent local IDE files from polluting the working tree.
  * Applied a non-breaking regex fix (`.replace(/\.js$/, '')`) to `scripts/build-indexes.mjs`. This allows the index generator to safely handle future directory-based test suites (like `es.temporal/duration/`) without inadvertently slicing folder names.
* **Centralized Harness Support:** Added `tests/helpers/temporal-helpers.js`, porting vital assertion macros (e.g., `assertDuration`) and ISO constants [from Test262](https://github.com/tc39/test262/blob/7f2001f611ef96532c68e000b413c5599df94147/harness/temporalHelpers.js). Assertions have been refactored to rely on passing the localized test framework context rather than assuming a leaked global context.
* **Global & Pure Baselines:** Created flat baseline files (`tests/unit-global/es.temporal.js` and `tests/unit-pure/es.temporal.js`) to verify the root namespace and basic descriptor intents (using safe target checks like `Object.isExtensible`).

### Architectural Guardrails Included

* Wrapped property descriptor evaluations inside `if (DESCRIPTORS)` environments to maintain compatibility with legacy testing engines.
* Pure tests contain active safety gates to gracefully skip execution until the relative entry points are generated in `core-js-pure`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ node_modules/
 *.swp
 .DS_Store
 .idea
+.vscode
 
 /bundles/
 /docs/web/blog/
--- a/scripts/build-indexes.mjs
+++ b/scripts/build-indexes.mjs
@@ -12,7 +12,7 @@ async function generateTestsIndex(name, pkg) {
   const files = await fs.readdir(dir);
   return fs.writeFile(`${ dir }/index.js`, `import '../helpers/qunit-helpers';\n\n${ files
     .filter(it => /^(?:es|esnext|helpers|web)\./.test(it))
-    .map(it => `import './${ it.slice(0, -3) }';\n`)
+    .map(it => `import './${ it.replace(/\.js$/, '') }';\n`)
     .join('') }${ pkg !== 'core-js' ? `\nimport core from '${ pkg }';\ncore.globalThis.core = core;\n` : '' }`);
 }
 
--- a/tests/helpers/temporal-helpers.js
+++ b/tests/helpers/temporal-helpers.js
@@ -1,0 +1,381 @@
+// This is a `core-js` friendly version of Test262's Temporal test helpers:
+// https://github.com/tc39/test262/blob/7f2001f611ef96532c68e000b413c5599df94147/harness/js
+// The above link is a copy of the `js` file at the time of writing
+
+import { GLOBAL } from './constants.js';
+
+const Symbol = GLOBAL.Symbol || {};
+
+const ASCII_IDENTIFIER = /^[$A-Z_a-z][\w$]*$/u;
+
+// eslint-disable-next-line no-unused-vars -- Helper function to be used in later test helpers
+function formatPropertyName(propertyKey, objectName = '') {
+  switch (typeof propertyKey) {
+    case 'symbol': {
+      if (Symbol.keyFor(propertyKey) !== undefined) {
+        return `${ objectName }[Symbol.for('${ Symbol.keyFor(propertyKey) }')]`;
+      }
+
+      const description = propertyKey.description || '';
+
+      if (description.startsWith('Symbol.')) {
+        return `${ objectName }[${ description }]`;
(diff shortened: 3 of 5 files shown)


Raymondo97:
I did want to mention that the reason I decided to put the tests into a separate directory under the `unit-pure` and `unit-global` is due to the vast number of test that would need to be added (200+). We can either create a flat structure for all the Temporal property's tests, or we can nest those as well, similar to how the [Test262 tests](https://github.com/tc39/test262/tree/7f2001f611ef96532c68e000b413c5599df94147/test/built-ins/Temporal) are structured.

Raymondo97:
The changes made here do not pass the ESLint rules. I will mark the PR as ready when it succeeds, and the corrections are pushed up.

Raymondo97:
> я провірив все працює

Interesting. According to [the ESLint](https://github.com/zloirock/core-js/blob/c97f42dd89a71dcfc6cac002f3efeb8575471e1f/tests/eslint/eslint.config.js) rules, it shouldn’t pass. Especially for the `unit-pure` test. The `temporal-helpers.js` are also problematic. So unless I have my environment setup incorrectly, running `npm run lint` should throw errors.

zloirock:
@stasyu2009-ux if you wanna say something, please, use English.

zloirock:
@Raymondo97 sorry for the delayed reply. I am now fully focused on other tasks, I'll take a look in the next few days.

Raymondo97:
@zloirock No worries at all! I see that you've been working incredibly hard on v4.

To be clear, this PR establishes the `temporal-helpers.js` infrastructure, the regex fix for the index builder, and one flat baseline test (`es.temporal.js`).

However, as I look at porting the rest of the Temporal tests from Test262, I'm realizing the sheer volume will be massive. Moving forward, I'm wondering if it may be beneficial to match Test262's layout. We could group the tests into sub-directories inside an `es.temporal/` folder. For example:

```text
tests/unit-global/
└── es.temporal/
    ├── index.js             // Hardcoded imports for sub-suites
    ├── duration/
    │   ├── index.js         
    │   ├── constructor.js
    │   └── ...
    ├── plain-date/
    │   ├── index.js
    │   └── ...
    └── instant/

```

In order to support this, we'd need to decide how to apply ESLint settings to it. By hardcoding the `index.js` files into each sub-directory, we would likely need to add `'tests/unit-@(global|pure)/es.temporal/index.js'` to the [ESLint ignores array](https://github.com/zloirock/core-js/blob/c97f42dd89a71dcfc6cac002f3efeb8575471e1f/tests/eslint/eslint.config.js#L2355). We could also relax the [naming conventions](https://github.com/zloirock/core-js/blob/c97f42dd89a71dcfc6cac002f3efeb8575471e1f/tests/eslint/eslint.config.js#L2570-L2578) for the files within the `es.temporal/` directories to avoid redundancy.

*(Note: to support importing folders like this in the future, I already included a small regex fix to `scripts/build-indexes.mjs` in this PR so the test generator doesn't strip characters off of folder names).*

Would this nested approach align with your vision for the v3/v4 architecture, or would you prefer a different layout before I start porting the rest of the Test262 suite? I'll hold off on porting the massive test suites until I hear your thoughts on the architecture. I'm happy to adjust as needed!

stasyu2009-ux (commented):
I've already faked everything in this little task

PR #1545, opened 2026-07-01
Throw TypeError for non-integral Infinity size in Iterator.prototype.chunks/windows

spokodev:
`Iterator.prototype.chunks` and `Iterator.prototype.windows` throw a `RangeError` instead of a `TypeError` when the size is a non-integral `+Infinity`/`-Infinity`.

### Repro

```js
require('core-js/actual/iterator');
[1, 2, 3].values().chunks(Infinity).next();   // RangeError, spec wants TypeError
[1, 2, 3].values().windows(-Infinity).next(); // RangeError, spec wants TypeError
```

### Root cause

The size validation used `Math.floor(chunkSize) !== chunkSize` as the integrality check. `Math.floor(Infinity) === Infinity`, so `+Infinity`/`-Infinity` pass that check and fall through to the range check `chunkSize >>> 0 !== chunkSize` (`Infinity >>> 0` is `0`), producing a `RangeError`.

### Fix

Use the canonical `isIntegralNumber` helper, which is finite-aware (`isFinite(it) && floor(it) === it`). `+Infinity`/`-Infinity` are not integral Numbers, so they now throw a `TypeError`.

### Authority

The Iterator Chunking proposal specifies, for both `chunks` and `windows`, throwing a `TypeError` when the size is not an integral Number before the `[1, 2^32-1]` `RangeError` check. ECMA-262 defines an integral Number as a finite value, so `+Infinity`/`-Infinity` are not integral and must throw a `TypeError`.

This matches the existing `2.1`/`NaN` `TypeError` behavior and leaves `-0`/`-1`/`2^32` on the `RangeError` path unchanged. Tests added to all four suites (global and pure, `chunks` and `windows`).

### DIFF
--- a/packages/core-js/internals/iterator-chunk-size-validation.js
+++ b/packages/core-js/internals/iterator-chunk-size-validation.js
@@ -1,14 +1,14 @@
 'use strict';
+var isIntegralNumber = require('../internals/is-integral-number');
 var iteratorClose = require('../internals/iterator-close');
 
 var $RangeError = RangeError;
 var $TypeError = TypeError;
-var floor = Math.floor;
 
 var CHUNK_SIZE_VALIDATION_ERROR = 'Chunk size must be an integer in [1, 2^32-1]';
 
 module.exports = function (chunkSize, iterator) {
-  if (typeof chunkSize != 'number' || floor(chunkSize) !== chunkSize) {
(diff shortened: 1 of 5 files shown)


zloirock (approved):

zloirock:
Thanks.

PR #1553, opened 2026-08-01
fix: correct Iterator zip entry exports

ShiroKSH:
Fix the direct `es/iterator/zip` and `es/iterator/zip-keyed` entries to export the static `Iterator` methods, rather than unbinding nonexistent prototype methods. Also make `actual/iterator/zip-keyed` load its matching stable entry.

Tests:
- `npm run prepare`
- `npm run test-entries`
- `npm run bundle`
- `npm run test-unit-node`
- `npm run check`


zloirock:
Thanks.

zloirock (approved):

